PR #1, opened 2016-09-06
Add script to import flow events

rfk:
A super-simple script to import the new "flow events" being exported as of https://bugzilla.mozilla.org/show_bug.cgi?id=1289474; @philbooth r?

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -1,0 +1,143 @@
+#
+# Script to import "flow event" metrics from S3 into redshift.
+#
+
+import os
+import json
+import time
+import datetime
+
+import postgres
+
+import boto.s3
+import boto.provider
+
+# Load config from disk,
+# and pull in credentials from the environment.
+
+with open("config.json") as f:
+    CONFIG = json.loads(f.read())
+
+if "aws_access_key_id" not in CONFIG:
+    p = boto.provider.Provider("aws")
+    CONFIG["aws_access_key_id"] = p.get_access_key()
+    CONFIG["aws_secret_access_key"] = p.get_secret_key()
+
+DB = "postgresql://{db_username}:{db_password}@{db_host}:{db_port}/{db_name}".format(**CONFIG)
+
+# Event data files are named like "flow-2016-02-15.csv"
+# and contain events for that day.
+
+EVENTS_BUCKET = "net-mozaws-prod-us-west-2-pipeline-analysis"
+EVENTS_PREFIX = "fxa-flow/data/"
+EVENTS_FILE_URL = "s3://" + EVENTS_BUCKET + "/" + EVENTS_PREFIX + "flow-{day}.csv"
+
+# We import each into is own table and maintain a UNION ALL view over them.
+
+Q_CREATE_TABLE = """
+    CREATE TABLE IF NOT EXISTS flow_events (
+      timestamp BIGINT NOT NULL sortkey, 
+      flowid VARCHAR(64) distkey,
+      type VARCHAR(30) NOT NULL,
+      flowTime INTEGER,
+      userAgentBrowser VARCHAR(40),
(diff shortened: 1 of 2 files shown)


import_flow_events.py
```diff
@@ -0,0 +1,143 @@
...
+
+Q_CREATE_TABLE = """
+    CREATE TABLE IF NOT EXISTS flow_events (
+      timestamp BIGINT NOT NULL sortkey, 
+      flowid VARCHAR(64) distkey,
+      type VARCHAR(30) NOT NULL,
+      flowTime INTEGER,
+      userAgentBrowser VARCHAR(40),
+      userAgentVersion VARCHAR(40),
+      userAgentOS VARCHAR(40),
```
philbooth, line 45: Whoops, I gave these three different names in the feature doc, will update there.

import_flow_events.py
```diff
@@ -0,0 +1,143 @@
...
+
+EVENTS_BUCKET = "net-mozaws-prod-us-west-2-pipeline-analysis"
+EVENTS_PREFIX = "fxa-flow/data/"
+EVENTS_FILE_URL = "s3://" + EVENTS_BUCKET + "/" + EVENTS_PREFIX + "flow-{day}.csv"
+
+# We import each into is own table and maintain a UNION ALL view over them.
+
+Q_CREATE_TABLE = """
+    CREATE TABLE IF NOT EXISTS flow_events (
+      timestamp BIGINT NOT NULL sortkey,
```
philbooth, line 39: Fwiw, I've found the `TIMESTAMP` and `INTERVAL` data types in postgres quite intuitive to work with and I like the explicitness of the type, saying exactly what the are. Is it mainly for consistency with the activity event tables that we can't use those here?

philbooth:
Ha ha, I spent much of Friday and yesterday [working on the same thing](https://github.com/mozilla/fxa-activity-metrics/compare/phil/import-new-events?expand=1) (although much slower than you, evidently). 😄

But it puts me in a good frame of mind to review this. The first thing I notice is that this doesn't split them into the separate `flow_metadata` and `flow_events` tables described by the feature doc. This was the sticking point I'd got as far as fwiw, I was copying into a big temporary table then doing `INSERT` queries from that into the respective final tables. (although I hadn't got it to set the aggregate properties like `duration` etc yet)

Anyway, couple of inline question but aside from those looks good. I notice there's no attempt to update the activity event tables here, but I'm assuming that's intentional. (think it was just the addition of `deviceId` and then changing comments and variable names to say `day` instead of `week`)

import_flow_events.py
rfk: It's mainly for "I couldn't be bothered changing it" reasons, which are pretty crappy reasons :-)

We've going to have to revisit how the activity event imports are handled anyways, since they are more frequent and include more data.  So I'll happily take your guidance on the right types etc here.  Might be easiest if you just do them as a PR to this PR?

rfk:
I also wouldn't be sad if we threw this one away and replaced it with yours, esp. if you're further along with the flow metadata stuff than I am here (which as you noticed, I'm currently at "zero percent")

philbooth:
> I also wouldn't be sad if we threw this one away and replaced it with yours, esp. if you're further along with the flow metadata stuff than I am here (which as you noticed, I'm currently at "zero percent")

Well I'll definitely carry on it with it this morning, either adding to your branch or bring stuff from yours in to mine. Not sure which yet, it will get done one way or another. 😄

rfk:
(as you may be able to tell, I'm pretty excited to see what we can start learning from these new events!)

philbooth:
> (as you may be able to tell, I'm pretty excited to see what we can start learning from these new events!)

If you want to just merge this right away so you can play with stuff, I'm happy to rebase. r+ in that case!

PR #2, opened 2016-09-07
WIP: Update redshift import scripts

philbooth:
@rfk, these are my import script changes finished off and rebased against latest master.

I've literally only just finished them, so I'm sure there's still heaps of syntax and logical errors. But I wanted to take advantage of the time difference to get early feedback on a few things, if you get the time to eyeball it.

I'll call out the items of interest inline (go alliteration!), but to summarise:
- `import_activity_events.py` is updated with `deviceId`.
- References to `week` in `import_activity_events.py` have been updated to `day`.
- `import_flow_events.py` splits the data into `flow_metadata` and `flow_events`.
- The flow tables use TIMESTAMP and INTERVAL rather than BIGINT (I think this makes queries and graphs nicer).
- It uses a temporary `flow_csv` table for importing the raw data from CSV file (conversions happen on subsequent `INSERT`).

More detail inline.


import_flow_events.py
```diff
@@ -32,17 +32,22 @@
...
+
+Q_CREATE_CSV_TABLE = """
+    CREATE TABLE IF NOT EXISTS flow_csv (
+      timestamp BIGINT NOT NULL SORTKEY,
       type VARCHAR(30) NOT NULL,
-      flowTime INTEGER,
-      userAgentBrowser VARCHAR(40),
-      userAgentVersion VARCHAR(40),
-      userAgentOS VARCHAR(40),
+      flowId VARCHAR(64) NOT NULL DISTKEY,
```
philbooth: I've set the order of this table to directly match the order of the CSV file. Figured that may save confusion down the line (or maybe that's just me, I got a bit confused before I did this earlier).

import_flow_events.py
```diff
@@ -54,22 +59,63 @@
...
+Q_CREATE_METADATA_TABLE = """
+    CREATE TABLE IF NOT EXISTS flow_metadata (
+      flowId VARCHAR(64) NOT NULL UNIQUE SORTKEY,
+      beginTime TIMESTAMP NOT NULL,
+      duration INTERVAL NOT NULL,
+      completed BOOLEAN NOT NULL,
+      newAccount BOOLEAN NOT NULL,
+      browser VARCHAR(40),
+      browserVersion VARCHAR(40),
+      os VARCHAR(40),
```
philbooth: I've renamed these fields because the whole `userAgent` prefix vexes me whenever I have to type it out. Figured that if it annoys me, it may annoy other people too. And the shorter names seem clear enough to me in the context.

import_flow_events.py
```diff
@@ -54,22 +59,63 @@
...
+      flowTime INTERVAL NOT NULL,
+      flowId VARCHAR(64) NOT NULL DISTKEY,
+      type VARCHAR(30) NOT NULL
+    );
+"""
 
-Q_CLEAR_EVENTS = """
+Q_CHECK_FOR_DAY = """
+    SELECT timestamp FROM flow_events
+    WHERE timestamp::date >= {day}
```
philbooth: Postgres' casting operator lets us ditch the `day_to_timestamp` function further down, because we can compare against the day string directly.

import_flow_events.py
```diff
@@ -54,22 +59,63 @@
...
+      utmCampaign VARCHAR(40),
+      utmContents VARCHAR(40),
+      utmMedium VARCHAR(40),
+      utmSource VARCHAR(40),
+      utmTerm VARCHAR(40)
+    );
+"""
+Q_CREATE_EVENTS_TABLE = """
+    CREATE TABLE IF NOT EXISTS flow_events (
+      timestamp TIMESTAMP NOT NULL SORTKEY,
```
philbooth, line 87: I've added `timestamp` to this table and made it `SORTKEY`, which is not in the spec. I use `timestamp` as the deletion criteria in `Q_CLEAR_DAY_EVENTS`.

I did briefly delete from that table using an `ON DELETE CASCADE` foreign key reference instead, but then I realised that was error prone in the cases where the `flow.begin` event occurs on a different day to the target event.

import_flow_events.py
```diff
@@ -84,18 +130,81 @@
...
+      utmCampaign,
+      utmContents,
+      utmMedium,
+      utmSource,
+      utmTerm
+    )
+    SELECT (
+      begin.flowId,
+      (begin.timestamp * 1000.0),
+      (durations.flowTime * 1000.0),
```
philbooth: The downside of using `TIMESTAMP` and `INTERVAL` is the need to do manual conversion like this on import, because the resolution for both is microseconds. But that feels like a price worth paying for the simpler math in queries and the automatic formatting in re:dash graphs.

import_flow_events.py
```diff
@@ -84,18 +130,81 @@
...
-Q_CHECK_FOR_DAY = """
-    SELECT timestamp FROM flow_events
-    WHERE timestamp >= {timestamp}
-    AND timestamp < {timestamp} + 86400
-    LIMIT 1;
+Q_INSERT_METADATA = """
+    WITH durations AS (
+      SELECT flowId, MAX(flowTime)
+      FROM flow_csv
+      GROUP BY flowId
```
philbooth: Haven't actually tried this query out in isolation yet, maybe it doesn't do what I think it does.

import_flow_events.py
```diff
@@ -84,18 +130,81 @@
...
+      begin.utmSource,
+      begin.utmTerm
+    )
+    FROM flow_csv AS begin
+    INNER JOIN durations
+      ON begin.flowId = durations.flowId AND begin.type = 'flow.begin'
+    LEFT JOIN flow_csv AS created
+      ON begin.flowId = created.flowId AND created.type = 'account.created'
+    LEFT JOIN flow_csv AS signed
+      ON begin.flowId = signed.flowId AND signed.type = 'account.signed';
```
philbooth: And I haven't tried this joining-against-yourself business out yet either. Will it work? Is it unconventional? Is there some other way that's better?

import_flow_events.py
```diff
@@ -54,22 +59,63 @@
...
+      flowId VARCHAR(64) NOT NULL DISTKEY,
+      type VARCHAR(30) NOT NULL
+    );
+"""
 
-Q_CLEAR_EVENTS = """
+Q_CHECK_FOR_DAY = """
+    SELECT timestamp FROM flow_events
+    WHERE timestamp::date >= {day}
+    AND timestamp::date < {day} + 1
```
philbooth: Oh, just realised this `+ 1` is wrong too, need to convert it to `DATE` first.

import_flow_events.py
rfk: OK; we could also consider `ua` as a shorter prefix, but I'm going to resist the temptation to bikeshed - pick whichever feels right to you.

We should consider updating the activity-events tables for consistency, and may need to adjust any existing re:dash graphs that use this columns.

import_flow_events.py
rfk: I don't recall whether redshift properly processes foreign keys and `ON DELETE CASCADE` anyways.  I do know that it has problems handling unique key constraints (and by "problems" I mean "it just ignores them and lets you insert duplicates").  So we're probably better off doing own own explicit management of these things.

import_flow_events.py
rfk: Yeah, I agree, having them as proper timestamp columns is the Right Thing even if it makes the import a bit weird.

import_activity_events.py
```diff
@@ -173,11 +176,11 @@ def import_events(force_reload=False):
...
 
 
-def fixup_event_data(b, week):
+def fixup_event_data(b, day):
     """Download a data file, remove invalid lines, and re-upload"""
-    orig_key = b.get_key(urlparse.urlparse(EVENTS_FILE_URL.format(week=week)).path[1:])
+    orig_key = b.get_key(urlparse.urlparse(EVENTS_FILE_URL.format(day=day)).path[1:])
     assert orig_key is not None
-    fixed_key = b.new_key(urlparse.urlparse(EVENTS_FILE_FIXED_URL.format(week=week)).path[1:])
+    fixed_key = b.new_key(urlparse.urlparse(EVENTS_FILE_FIXED_URL.format(day=day)).path[1:])
```
rfk, line 183: These activity-event changes all look fine, but I suspect we cant run them as-is because it will mix the old and new data together in a single table.  Should we bite the bullet and migrate to a new table (say "activity_events") for the new 100%-sample-rate data, and keep it separate from the old sampled data?  (We could also move off using the union-of-views trick at the same time).

import_flow_events.py
```diff
@@ -54,22 +59,63 @@
       utmTerm VARCHAR(40)
     );
 """
+Q_CREATE_METADATA_TABLE = """
+    CREATE TABLE IF NOT EXISTS flow_metadata (
+      flowId VARCHAR(64) NOT NULL UNIQUE SORTKEY,
+      beginTime TIMESTAMP NOT NULL,
```
rfk: What's the advantage of using`flowId` for the `sortkey`?  An alternative might be to make it the `distkey`, and use `beginTime` as the `sortkey`, to let us to time-based summary queries from this table. (like "number of compelete flows per day, or time").

import_flow_events.py
rfk: (IOW, what you've done here sounds right to me)

import_flow_events.py
```diff
@@ -54,22 +59,63 @@
...
+    DELETE FROM flow_metadata
+    WHERE beginTime::date >= {day}::date
+    AND beginTime::date < {day}::date + 1;
+"""
+Q_CLEAR_DAY_EVENTS = """
     DELETE FROM flow_events
-    WHERE timestamp >= {timestamp}
-    AND timestamp < {timestamp} + 86400
+    WHERE timestamp::date >= {day}::date
+    AND timestamp::date < {day}::date + 1;
```
rfk: This does read much, much clearer than `86400` :-)

import_flow_events.py
```diff
@@ -84,18 +130,81 @@
...
+    FROM flow_csv AS begin
+    INNER JOIN durations
+      ON begin.flowId = durations.flowId AND begin.type = 'flow.begin'
+    LEFT JOIN flow_csv AS created
+      ON begin.flowId = created.flowId AND created.type = 'account.created'
+    LEFT JOIN flow_csv AS signed
+      ON begin.flowId = signed.flowId AND signed.type = 'account.signed';
+"""
+Q_INSERT_EVENTS = """
+    INSERT INTO flow_metadata (
```
rfk: s/flow_metadata/flow_events/

import_flow_events.py
```diff
@@ -32,17 +32,22 @@
 EVENTS_PREFIX = "fxa-flow/data/"
 EVENTS_FILE_URL = "s3://" + EVENTS_BUCKET + "/" + EVENTS_PREFIX + "flow-{day}.csv"
 
-# We import each into is own table and maintain a UNION ALL view over them.
+# There are three tables:
+#   * flow_csv      - raw data from the CSV file
```
rfk: One nit, I wonder if we can name this table so that it's more obviously a temporary import table.  `raw_flow_data_import` or something like that?  No biggie but something to think about.

import_flow_events.py
rfk: (also please put a comment as to what the multiplication is doing here, in the SQL itself)

import_flow_events.py
rfk: If the joins here get too unwieldy, another option could be to update the table piecemeal, along the lines of:
1. Insert a new row for each `flow.begin` event in the import table, defaulting to existing-account, incomplete, etc.
2. Do an update to set the duration for each flowid in the import table.
3. Do an update to set `newAccount=1` for all `account.created` events in the import table.
4. Do an update to set `completed=1` for all `account.signed` events in the import table.

This might make it easier to handle flows that cross the boundary between days, e.g. if the `flow.begin` event is at 11:59 on the 1st of May, and the `account.signed` event is at 12:03 on the 2nd of May.  It would start off being an incomplete flow, but get updated to its correct status when the next day's data is processed.

On the downside, it would leave more room for something going wrong between each of the steps.

rfk:
@philbooth I like where this is heading, good stuff!

import_activity_events.py
philbooth: > Should we bite the bullet and migrate to a new table (say "activity_events") for the new 100%-sample-rate data, and keep it separate from the old sampled data?

Absolutely. I should have mentioned, I haven't touched the activity event script since our 1:1 the other day, so none of that stuff is taken into account yet. For the record, my recollection is single table, then make sure we're happy with that schema, then ask Wes about the possibility of retro-exporting old data at the new frequency/sample rate and with an empty device id.

I'll make sure I bring this script up to date today.

import_flow_events.py
philbooth: This is probably just me not understanding `DISTKEY` correctly yet. In my mental model, because `flowId` is unique here, it was the wrong choice. I thought the `DISTKEY` should be something for which there was more than one row of each. I'll go and RTFM again. 😄

import_flow_events.py
philbooth: > we could also consider `ua` as a shorter prefix

`ua` is tolerable, I'll change.

> We should consider updating the activity-events tables for consistency, and may need to adjust any existing re:dash graphs that use this columns.

Yep, noted.

import_flow_events.py
philbooth: > This might make it easier to handle flows that cross the boundary between days

Excellent point, I hadn't thought that scenario through properly at all. Will change to updates.

import_flow_events.py
rfk: Hmm, yes I think you're right, there's not a lot of point having distkey be a unique column.  OTOH, I still feel like we're more likely to want query in a timestamp-driven fashion.

import_flow_events.py
```diff
@@ -54,22 +59,63 @@
       utmTerm VARCHAR(40)
     );
 """
+Q_CREATE_METADATA_TABLE = """
+    CREATE TABLE IF NOT EXISTS flow_metadata (
+      flowId VARCHAR(64) NOT NULL UNIQUE,
+      beginTime TIMESTAMP NOT NULL SORTKEY,
+      duration INTERVAL NOT NULL,
```
philbooth: _Sigh_, [`INTERVAL` is not supported in redshift](http://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-datatypes.html).

philbooth:
So, I dropped `flow_events` and took a stab at re-importing the flow events using the updated `import_flow_events.py`. As things stand currently, it is failing with:

```
  File "import_flow_events.py", line 260, in <module>
    import_events(True)
  File "import_flow_events.py", line 245, in import_events
    db.run(Q_INSERT_METADATA)
    cursor.run(sql, parameters)
    self.execute(sql, parameters)
psycopg2.ProgrammingError: syntax error at or near "INSERT"
LINE 7:     INSERT INTO flow_metadata (
```

Not sure what the problem is yet, maybe something to do with the preceding `WITH`?

import_flow_events.py
```diff
@@ -118,26 +265,31 @@ def import_events(force_reload=False):
...
+            db.run(Q_INSERT_EVENTS)
+            db.run(Q_UPDATE_TIMESTAMP)
 
         # Print the timestamps for sanity-checking.
-        print "MIN TIMESTAMP", db.one("SELECT MIN(timestamp) FROM flow_events")
-        print "MAX TIMESTAMP", db.one("SELECT MAX(timestamp) FROM flow_events")
+        print "MIN TIMESTAMP", db.one("SELECT MIN(timestamp) FROM temporary_raw_flow_data")
+        print "MAX TIMESTAMP", db.one("SELECT MAX(timestamp) FROM temporary_raw_flow_data")
+
+        db.run(Q_DROP_CSV_TABLE)
```
rfk: Should we drop the import table after each iteration of the loop, rather than once at the end?  Otherwise ISTM that the `Q_UPDATE_BEGIN_TIME`, `Q_UPDATE_DURATION` etc could be operating on the same data multiple times.

philbooth: Whoops, you are absolutely right of course.

philbooth:
Going to break up the activity events and flow events as separate PRs, closing this one.

PR #3, opened 2016-09-11
Normalise flow event metadata

philbooth:
@rfk, this is just the changes for the flow event import script.

These now import cleanly into redshift, having fixed all of my stupid syntax errors and typos that were getting in the way last week. I imported the events again just now so there are 9 days of event data available in redshift with this schema, if you want to have a play around.

One of the many things I learned in getting it to work was that redshift automatically lower-cases identifiers in queries, so I've switched from camelCase to snake_case for the column names. Not sure how that sits with you stylistically, I'm happy to change if you find it abhorrent.

Tested both with and without the `force_reload` flag set to `True`.

Sample log output:

```
net-mozaws-prod-us-west-2-pipeline-analysis fxa-flow/data/
2016-09-02
2016-09-03
2016-09-04
2016-09-05
2016-09-06
2016-09-07
2016-09-08
2016-09-09
2016-09-10
LOADING 9 DAYS OF DATA
LOADING 2016-09-10
  MIN TIMESTAMP 1473465600
  MAX TIMESTAMP 1473551999
LOADING 2016-09-09
  MIN TIMESTAMP 1473379200
  MAX TIMESTAMP 1473465599
LOADING 2016-09-08
  MIN TIMESTAMP 1473292800
  MAX TIMESTAMP 1473379199
LOADING 2016-09-07
  MIN TIMESTAMP 1473206400
  MAX TIMESTAMP 1473292799
LOADING 2016-09-06
  MIN TIMESTAMP 1473120000
  MAX TIMESTAMP 1473206399
LOADING 2016-09-05
  MIN TIMESTAMP 1473033600
  MAX TIMESTAMP 1473119999
LOADING 2016-09-04
  MIN TIMESTAMP 1472947200
  MAX TIMESTAMP 1473033599
LOADING 2016-09-03
  MIN TIMESTAMP 1472860800
  MAX TIMESTAMP 1472947199
LOADING 2016-09-02
  MIN TIMESTAMP 1472774660
  MAX TIMESTAMP 1472860799
```

Can I get an r?

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -269,4 +269,4 @@ def import_events(force_reload=False):
         db.run("COMMIT TRANSACTION")
 
 if __name__ == "__main__":
-    import_events(True)
+    import_events(False)
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -32,74 +32,197 @@ EVENTS_BUCKET = "net-mozaws-prod-us-west-2-pipeline-analysis"
 EVENTS_PREFIX = "fxa-flow/data/"
 EVENTS_FILE_URL = "s3://" + EVENTS_BUCKET + "/" + EVENTS_PREFIX + "flow-{day}.csv"
 
-# We import each into is own table and maintain a UNION ALL view over them.
+# There are three tables:
+#   * temporary_raw_flow_data - raw data from the CSV file
+#   * flow_metadata           - metadata for each flow
+#   * flow_events             - individual flow events
 
-Q_CREATE_TABLE = """
-    CREATE TABLE IF NOT EXISTS flow_events (
-      timestamp BIGINT NOT NULL sortkey, 
-      flowid VARCHAR(64) distkey,
+Q_DROP_CSV_TABLE = "DROP TABLE IF EXISTS temporary_raw_flow_data;"
+
+Q_CREATE_CSV_TABLE = """
+    CREATE TABLE IF NOT EXISTS temporary_raw_flow_data (
+      timestamp BIGINT NOT NULL SORTKEY,
       type VARCHAR(30) NOT NULL,
-      flowTime INTEGER,
-      userAgentBrowser VARCHAR(40),
-      userAgentVersion VARCHAR(40),
-      userAgentOS VARCHAR(40),
+      flow_id VARCHAR(64) NOT NULL DISTKEY,
+      flow_time BIGINT NOT NULL,
+      ua_browser VARCHAR(40),
+      ua_version VARCHAR(40),
+      ua_os VARCHAR(40),
       context VARCHAR(40),
       entrypoint VARCHAR(40),
       migration VARCHAR(40),
       service VARCHAR(40),
-      utmCampaign VARCHAR(40),
-      utmContent VARCHAR(40),
-      utmMedium VARCHAR(40),
-      utmSource VARCHAR(40),
-      utmTerm VARCHAR(40)
+      utm_campaign VARCHAR(40),
+      utm_content VARCHAR(40),
+      utm_medium VARCHAR(40),
+      utm_source VARCHAR(40),
+      utm_term VARCHAR(40)
+    );
+"""
+Q_CREATE_METADATA_TABLE = """
+    CREATE TABLE IF NOT EXISTS flow_metadata (
+      flow_id VARCHAR(64) NOT NULL UNIQUE,
+      begin_time TIMESTAMP NOT NULL SORTKEY,
+      -- Ideally duration would be type INTERVAL
+      -- but redshift doesn't support that.
+      duration INTEGER NOT NULL DEFAULT 0,
+      completed BOOLEAN NOT NULL DEFAULT FALSE,
+      new_account BOOLEAN NOT NULL DEFAULT FALSE,
+      ua_browser VARCHAR(40),
+      ua_version VARCHAR(40),
+      ua_os VARCHAR(40),
+      context VARCHAR(40),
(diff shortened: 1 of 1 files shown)


philbooth:
Hmm, after seeing some oddities in my bigger queries, I decided to run some more basic sanity checks.

In particular, it looks like there is something wrong with the `service` parameter because it appears to be empty for every row. I'll dig deeper into this tomorrow, but let me know if you see anything I've done wrong.

Full transcript:

``` sql
SELECT COUNT(*)
FROM flow_events;
-- 14,190,568

SELECT COUNT(*)
FROM flow_events
WHERE type = 'flow.begin';
-- 10,069,178

SELECT COUNT(*)
FROM flow_events
WHERE type = 'account.created';
-- 508,181

SELECT COUNT(*)
FROM flow_events
WHERE type = 'account.login';
-- 566,693

SELECT COUNT(*)
FROM flow_events
WHERE type = 'account.signed';
-- 740,900

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id;
-- 14,175,526

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.service != '';
-- 0

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.duration = 0;
-- 9,049,876

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.duration > 0;
-- 5,125,650

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.completed = FALSE;
-- 9,975,154

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.completed = TRUE;
-- 4,200,372

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.new_account = FALSE;
-- 12,016,305

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.new_account = TRUE;
-- 2,159,221
```

(the discrepancy of events that have no metadata I assume is just trailing events where the `flow.begin` was emitted before logs began)

rfk:
> In particular, it looks like there is something wrong with the service parameter because
> it appears to be empty for every row.

Looking at the raw data, it seems the `flow.begin` event has empty context, service, entrypoint etc parameters.  Since the `flow_metadata` table is only populated from this event type, that's probably why `service` is missing.

I guess this is a bug with the `flow.begin` event itself.

rfk:
Until the flow.begin events start reporting this data, one option is to try to get it off the other events in the flow.  You could try something like:

```
UPDATE flow_metadata
 SET
    service = COALESCE(service, metricsContext.service),
    entrypoint = COALESCE(entrypoint, metricsContext.entrypoint)
FROM (
  SELECT flow_id,
     MAX(service) AS service,
     MAX(entrypoint) as entrypoint
  FROM temporary_raw_flow_data
  GROUP BY flow_id
 ) AS metricsContext
WHERE flow_metadata.flow_id = metricsContext.flow_id;
```

Which I think should select the non-null value from any of the events in the flow.

rfk:
@philbooth that aside, this is looking good to me, r+!

philbooth:
@rfk, I tweaked your recommended fix for the missing metrics context fields to use `CASE` rather than `COALESCE`, because the values are the empty string rather than `NULL`.

Works well:

``` sql
SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.context <> '';
-- 5,340,744

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.entrypoint <> '';
-- 5,186,692

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.migration <> '';
-- 9,446

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.service <> '';
-- 5,650,980

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.utm_campaign <> '';
-- 2,491,735

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.utm_content <> '';
-- 2,491,735

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.utm_medium <> '';
-- 2,463,977

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.utm_source <> '';
-- 2,463,977

SELECT COUNT(*)
FROM flow_events AS e
INNER JOIN flow_metadata AS m
ON e.flow_id = m.flow_id
WHERE m.utm_term <> '';
-- 145
```

PR #4, opened 2016-09-12
feat(scripts): insert all activity events into one big table

philbooth:
@rfk, this inserts the activity events with the following changes to the schema
- All the events are inserted into one table, called `activity_events`.
- The `device_id` column is added.
- `timestamp` is now a `TIMESTAMP` rather than a `BIGINT`.

The import is running at the moment, but I've watched the first few days go in without failing.

r?

### DIFF
--- a/import_activity_events.py
+++ b/import_activity_events.py
@@ -28,7 +28,7 @@ if "aws_access_key_id" not in CONFIG:
 DB = "postgresql://{db_username}:{db_password}@{db_host}:{db_port}/{db_name}".format(**CONFIG)
 
 # Event data files are named like "events-2016-02-15.csv"
-# and contain events for the week starting that date.
+# and contain events for the specified date.
 # Unfortunately some of the data files have missing fields :-(
 # We work around this by downloading the file, fixing it up,
 # and uploading the fixed copy to a new location.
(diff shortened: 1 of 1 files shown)


philbooth:
Putting the `WIP` label on this and de-asking for review, sorry.

It still gets into a pickle when it hits the old CSV files (obviously). I've commented out all the fix-up stuff for now so that the import completes and we can play with the data in re:dash.

But longer term, I'm not sure whether the correct fix is to remove the old files from S3 or instead distinguish them in the script and ignore. Maybe based on the field count or something.

Let me know what you think @rfk.

rfk:
> It still gets into a pickle when it hits the old CSV files

In retrospect, we should have asked for the new-format data to be exported to a new location.  Oh well.  If we plan to ask for a back-fill anyway, maybe we can ask for an updated location as part of it.

Perhaps the easiest fix in the meantime is to just put a stopper in the script here that says "don't try to import events older than X date".

rfk:
That issue aside, r+ on the changes here.

rfk:
Also we should flag @davismtl about what's happening here. 

Alex, due to the addition of new events and new fields in our event data exports, we're having to rejigger the way events are imported into redash.  The end result might be that we have to re-do some of existing dashboards to support the new event and table formats, but we'll need to wait for the dust to settle here before it's clear what the impact will be.

philbooth:
> Perhaps the easiest fix in the meantime is to just put a stopper in the script here that says "don't try to import events older than X date".

Good idea, done this [here](https://github.com/mozilla/fxa-activity-metrics/pull/4/files#diff-eed61535049fc88e2eee0fe5cc92d3d7R136).

philbooth:
Just tested again, with and without `force_reload`. Merging.

davismtl:
No worries if we have to re-do a few dashboards. As long as it's for the best. :)

PR #5, opened 2016-09-13
fix(scripts): calculate daily summaries against new schema

philbooth:
Updated schema for daily summaries, available to play with now in re:dash and enabled in crontab. Tested both with and without the `day_from` and `day_until` arguments set.

@rfk r?

### DIFF
--- a/calculate_daily_summary.py
+++ b/calculate_daily_summary.py
@@ -22,31 +22,30 @@ with open("config.json") as f:
 DB = "postgresql://{db_username}:{db_password}@{db_host}:{db_port}/{db_name}".format(**CONFIG)
 
 # For the daily device activity summary,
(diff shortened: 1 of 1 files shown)


rfk:
LGTM, r+

philbooth:
For the record, I also just did a quick test to check that everything ran okay from cron overnight. Looks fine:

``` sql
SELECT MAX(begin_time) FROM flow_metadata;
-- 13/09/16 23:59

SELECT MAX(timestamp) FROM flow_events;
-- 13/09/16 23:59

SELECT MAX(timestamp) FROM activity_events;
-- 13/09/16 23:59

SELECT MAX(day) FROM daily_activity_per_device;
-- 13/09/16

SELECT MAX(day) FROM daily_multi_device_users;
-- 13/09/16
```

Issue #9, opened 2016-10-05
Use column compression


rfk:
We can probably save a lot of disk space by ensuring that we use column compression:

 http://docs.aws.amazon.com/redshift/latest/dg/t_Compressing_data_on_disk.html

Unfortunately it might require dropping and re-creating our tables, but it will be interesting to do some analysis and see if the gain is worth it.  At the very least, we should decide what compression would be useful for when we re-create these tables under Ops control.

shane-tomlinson:
@rfk - can you triage this?

rfk:
Redshift has a thing that will suggest compression schemes based on existing table data, here's what it says for ours:

```
analyze compression activity_events
fxa-# ;
      Table      |   Column   | Encoding 
-----------------+------------+----------
 activity_events | timestamp  | lzo
 activity_events | uid        | lzo
 activity_events | type       | lzo
 activity_events | device_id  | lzo
 activity_events | service    | lzo
 activity_events | ua_browser | lzo
 activity_events | ua_version | lzo
 activity_events | ua_os      | bytedict
(8 rows)
```

TBH I'm surprised it didn't recommend delta-compression for the timestamp columns, but _shrug_

rfk:
I copied our existing activity-events table into a compressed version, and it made a pretty big difference:

```
 fxa=# SELECT svv_table_info.table, encoded, size, pct_used, empty, unsorted, tbl_rows FROM svv_table_info;

             table               | encoded |  size  | pct_used | empty | unsorted |  tbl_rows  
---------------------------------+---------+--------+----------+-------+----------+------------
compressed_activity_events       | Y       | 518694 |   9.0697 |     0 |     0.00 | 5835574900
activity_events                  | N       | 885298 |  15.4801 |     0 |     3.38 | 5835574900
```

It also seems to have dramatically improved performance of scanning the table:
- `select min(timestamp) from compressed_activity_events` -> about 40 seconds
- `select min(timestamp) from activity_events` -> several minutes

So I'm going to go ahead and compress the existing tables, and have made a PR to record the change for future use: https://github.com/mozilla/fxa-activity-metrics/issues/13

rfk:
I've updated activity-event and flow-event tables to use compression.  Since we're planning on changing the schema for the rollup tables anyway, we can convert them to use compression as part of that work.

rfk:
We're now doing this on all tables, closing

PR #13, opened 2016-10-20
feat(compression): Use column compression in activity_events table.

rfk:
Connects to #9; @philbooth r?

I've done this on the existing DB by copying into a new table with compression enabled, and renaming it to replace the old table.

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -61,35 +61,35 @@ Q_CREATE_CSV_TABLE = """
 """
 Q_CREATE_METADATA_TABLE = """
     CREATE TABLE IF NOT EXISTS flow_metadata (
-      flow_id VARCHAR(64) NOT NULL UNIQUE,
-      begin_time TIMESTAMP NOT NULL SORTKEY,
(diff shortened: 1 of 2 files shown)


import_flow_events.py
```diff
@@ -61,35 +61,35 @@
...
-      migration VARCHAR(40),
-      service VARCHAR(40),
-      utm_campaign VARCHAR(40),
-      utm_content VARCHAR(40),
-      utm_medium VARCHAR(40),
-      utm_source VARCHAR(40),
-      utm_term VARCHAR(40)
+      duration INTEGER NOT NULL DEFAULT 0 ENCODE lzo,
+      completed BOOLEAN NOT NULL DEFAULT FALSE ENCODE raw,
+      new_account BOOLEAN NOT NULL DEFAULT FALSE ENCODE raw,
```
rfk, line 70: It's pretty hard to compress randomly-interleaved single-bit booleans, hence `raw` :-)

philbooth:
👍

PR #14, opened 2016-10-25
feat(devices): Include more device metadata in daily device activity rollup

### LINKED ISSUE #7: Tweak daily summary tables to enable engagement ratio segmentation
The [feature doc](https://github.com/mozilla/fxa/tree/master/features/FxA-70-kpi-dashboards#engagement-ratio) shows the engagement ratio segmented by desktop users, Android users and iOS users. Unless there is some special query magic we can use instead, we'll need to ~~create daily summary tables for these metrics, as we already do for multi-device users~~ add extra columns to the daily summary table for these metrics.

### LINKED ISSUE #12: Change mutli-device definition period from 5 days to 7 days
As suggested by @davismtl, we should define multi-device as "used a second device in the previous 7 days", rather than the current "previous 5 days".  If we're going to re-create the rollup tables anyway, that's probably a good opportunity to do this.

rfk:
Fixes #7, and Fixes #12 

@philbooth like we discussed last week, this would just include more device meta-data in the daily activity rollup.  Will it be enough to enable the required segmentation?

It also adds table compression, and changes the multi-device period to 7 days.

### DIFF
--- a/calculate_daily_summary.py
+++ b/calculate_daily_summary.py
@@ -26,9 +26,13 @@ DB = "postgresql://{db_username}:{db_password}@{db_host}:{db_port}/{db_name}".fo
 
 Q_DAILY_DEVICES_CREATE_TABLE = """
     CREATE TABLE IF NOT EXISTS daily_activity_per_device (
-      day DATE NOT NULL SORTKEY,
-      uid VARCHAR(64) NOT NULL DISTKEY,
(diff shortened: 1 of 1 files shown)


philbooth:
Splendid!

philbooth:
Fyi, I dropped the summary tables and kicked off a fresh `make summarize` just now.

Issue #16, opened 2016-10-26
"Value out of range for 4 bytes" error reported during flow event import


rfk:
Checking the mail spool for the ec2-user on our import box, I see a bunch of tracebacks like the following (this one was from today)

```
make: Entering directory `/home/ec2-user/fxa-activity-metrics'
./build/bin/python ./import_activity_events.py
LOADING 7 DAYS OF DATA
LOADING 2016-10-25
  MIN TIMESTAMP 1477353600
  MAX TIMESTAMP 1477439999
LOADING 2016-10-24
  MIN TIMESTAMP 1477267200
  MAX TIMESTAMP 1477353599
LOADING 2016-10-23
  MIN TIMESTAMP 1477180800
  MAX TIMESTAMP 1477267199
LOADING 2016-10-22
  MIN TIMESTAMP 1477094400
  MAX TIMESTAMP 1477180799
LOADING 2016-10-21
  MIN TIMESTAMP 1477008000
  MAX TIMESTAMP 1477094399
LOADING 2016-10-20
  MIN TIMESTAMP 1476921600
  MAX TIMESTAMP 1477007999
LOADING 2016-10-19
  MIN TIMESTAMP 1476835200
  MAX TIMESTAMP 1476921599
./build/bin/python ./import_flow_events.py
net-mozaws-prod-us-west-2-pipeline-analysis fxa-flow/data/
2016-09-02
2016-09-03
2016-09-04
2016-09-05
2016-09-06
2016-09-07
2016-09-08
2016-09-09
2016-09-10
2016-09-11
2016-09-12
2016-09-13
2016-09-14
2016-09-15
2016-09-16
2016-09-17
2016-09-18
2016-09-19
2016-09-20
2016-09-21
2016-09-22
2016-09-23
2016-09-24
2016-09-25
2016-09-26
2016-09-27
2016-09-28
2016-09-29
2016-09-30
2016-10-01
2016-10-02
2016-10-03
2016-10-04
2016-10-05
2016-10-06
2016-10-07
2016-10-08
2016-10-09
2016-10-10
2016-10-11
2016-10-12
2016-10-13
2016-10-14
2016-10-15
2016-10-16
2016-10-24
2016-10-25
2016-10-26
2016-11-14
LOADING 16 DAYS OF DATA
LOADING 2016-11-14
Traceback (most recent call last):
  File "./import_flow_events.py", line 303, in <module>
    import_events(False)
[...]

PR #17, opened 2016-10-27
fix(db): eliminate schema mismatches

philbooth:
Fixes the discrepancy you pointed out in https://github.com/mozilla/fxa-activity-metrics/issues/16#issuecomment-256288256, @rfk.

I figured it was better to bring these two up in length, rather than bring the temporary table down, for consistency with the lengths specified on the `flow_metadata` table.

And also bigger is better, right? 💪

r?

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -65,7 +65,7 @@ Q_CREATE_METADATA_TABLE = """
       begin_time TIMESTAMP NOT NULL SORTKEY ENCODE lzo,
       -- Ideally duration would be type INTERVAL
       -- but redshift doesn't support that.
-      duration INTEGER NOT NULL DEFAULT 0 ENCODE lzo,
+      duration BIGINT NOT NULL DEFAULT 0 ENCODE lzo,
       completed BOOLEAN NOT NULL DEFAULT FALSE ENCODE raw,
       new_account BOOLEAN NOT NULL DEFAULT FALSE ENCODE raw,
       ua_browser VARCHAR(40) ENCODE lzo,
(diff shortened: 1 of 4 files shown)


philbooth:
I've also taken the liberty of adding the retrospective import script in a separate commit.

Previously I thought that was a one-time gig but, if we're going to recreate these tables, we'll need to run that import again too. May as well keep it in the repo in case we need it again in the future.

philbooth:
And I just pushed _yet another_ commit, fixing the `flow_time` mismatch mentioned in https://github.com/mozilla/fxa-activity-metrics/issues/16#issuecomment-256658141. Pretty sure that's all of them now.

philbooth:
(you can squash all these btw)

import_flow_events.py
```diff
@@ -44,7 +44,7 @@
       timestamp BIGINT NOT NULL SORTKEY,
       type VARCHAR(30) NOT NULL,
       flow_id VARCHAR(64) NOT NULL DISTKEY,
-      flow_time BIGINT NOT NULL,
+      flow_time INTEGER NOT NULL,
```
seanmonstar: Sorry, may be missing context.

If `flow_time` is to be storing the timestamp from JavaScript (mentioned in https://github.com/mozilla/fxa-activity-metrics/issues/16#issuecomment-256658141), then 4 bytes is probably too small.

``` sh
$ node -p 'Date.now()'
1477585871454
$ node -p 'Math.pow(2, 32)'
4294967296
```

4294967296 would be the biggest number of milliseconds you can fit in a 4-byte `INTEGER`, if it is unsigned, half if it is signed. Assuming `BIGINT` is 8 bytes, that seems better?

philbooth: It's a time difference, for which the maximum value should be four hours (based on current config). If it overflows to `BIGINT` territory, it points at something wrong (which is how we spotted this fwiw, see https://github.com/mozilla/fxa-activity-metrics/issues/16#issuecomment-256658141).

Having said that, maybe it's possible that in the future there will be no limits to `flow_time`. In which case, switching to `BIGINT` now would be prudent.

@rfk, what do you reckon?

seanmonstar: Ah, so it captures how long the flow took? As you say, if the config is adjusted, it doesn't sound that impossible for a flow to overflow 4 bytes. It overflows after around 49 hours, so if a flow was able to take longer than 2 days, it won't fit.

philbooth: Yeah, you're right. Updated to `BIGINT`, thanks for the feedback @seanmonstar!

seanmonstar: I goofed. It overflows after 49 days, not 49 hours.

PR #22, opened 2016-11-07
fix(db): increase VARCHAR length for flow_events.type

### LINKED ISSUE #21: flow_events.type column is too short
Flow event imports are failing with this error:

```
  File "./import_flow_events.py", line 313, in <module>
    import_events(False)
  File "./import_flow_events.py", line 289, in import_events
    **CONFIG
    cursor.run(sql, parameters)
    self.execute(sql, parameters)
psycopg2.InternalError: Load into table 'temporary_raw_flow_data' failed.  Check 'stl_load_errors' system table for details.
```

And when I look up the message in the `stl_load_errors` table it says `String length exceeds DDL length` for the value `account.login.confirmedUnblockCode`.

philbooth:
Fixes #21, the new flow event type `account.login.confirmedUnblockCode` exceeds our defined length.

From what I can make out, you can't `ALTER COLUMN` directly in redshift. I guess I could rename the table then copy everything into a new table with the correct column definition, but it seems just as easy to drop the tables then recreate them tbh. We don't have all the historical flow event data like we do for activity events so it won't take long.

Anyway, either way, this script needs updating, I figure a length of 64 characters should be plenty for `type`?

@rfk r?

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -42,7 +42,7 @@ Q_DROP_CSV_TABLE = "DROP TABLE IF EXISTS temporary_raw_flow_data;"
 Q_CREATE_CSV_TABLE = """
     CREATE TABLE IF NOT EXISTS temporary_raw_flow_data (
       timestamp BIGINT NOT NULL SORTKEY,
-      type VARCHAR(30) NOT NULL,
+      type VARCHAR(64) NOT NULL,
       flow_id VARCHAR(64) NOT NULL DISTKEY,
       flow_time BIGINT NOT NULL,
       ua_browser VARCHAR(40),
@@ -89,7 +89,7 @@ Q_CREATE_EVENTS_TABLE = """
       -- but redshift doesn't support that.
       flow_time BIGINT NOT NULL ENCODE lzo,
       flow_id VARCHAR(64) NOT NULL DISTKEY ENCODE lzo,
-      type VARCHAR(30) NOT NULL ENCODE lzo
+      type VARCHAR(64) NOT NULL ENCODE lzo
     );
 """
 


rfk:
> I figure a length of 64 characters should be plenty for type?

Yes, this should be enough for anybody! r+

Do you think it would be worth asserting some of these constrains in the auth-server code, to guard against similar problems in future?

rfk:
I updated the column definition on the live db by doing:
- make a new column "bigger_type"
- update that using values from the existing "type" column
- drop the existing "type" column
- rename "bigger_type" to "type"

But will we need to re-import the data anyway in order to pick up the events with the longer names?

rfk:
> But will we need to re-import the data anyway in order to pick up the events with the longer names?

Nope, looks like all the existing events have short enough names, and I guess postgres doesn't do the helpful mysql thing of just truncating value that don't fit:

```
fxa=# select distinct type, length(type) from flow_events;
           type           | length 
--------------------------+--------
 flow.begin               |     10
 account.verified         |     16
 account.confirmed        |     17
 flow.signup.submit       |     18
 flow.oauth.signin.engage |     24
 flow.signin.submit       |     18
 flow.signin.engage       |     18
 flow.force-auth.engage   |     22
 flow.signup.begin        |     17
 flow.signin.begin        |     17
 flow.oauth.signup.begin  |     23
 account.created          |     15
 device.created           |     14
 account.keyfetch         |     16
 account.signed           |     14
 account.login            |     13
 customs.blocked          |     15
 flow.signup.engage       |     18
 flow.force-auth.submit   |     22
 flow.oauth.signup.engage |     24
 flow.oauth.signin.begin  |     23
 flow.have-account        |     17
 flow.force-auth.begin    |     21
(23 rows)
```

philbooth:
> ...will we need to re-import the data anyway in order to pick up the events with the longer names?

We won't need to re-import anything, because these transactions failed. There's a long-ish backlog of CSVs that remain unprocessed is all. The next import job _should_ just magically stop failing and bring all the missing data in.

philbooth:
> Nope, looks like all the existing events have short enough names, and I guess postgres doesn't do the helpful mysql thing of just truncating value that don't fit...

Never mind, beat me to it. 😄

Issue #20, opened 2016-11-04
Improve performance of segmented KPI queries


philbooth:
I updated the queries for the KPI dashboard to segment users by service and operating system. Quelle surprise, it made everything slower.

The time-to-device-connection query was okay, the execution time only increased to 12 minutes (but also, this query uses flow events, for which we have significantly fewer records). However, the engagement ratio and multi-device user queries became unusable. They were still running when I checked on them after 12 hours and then when I came back to them this morning, they'd failed with an unhelpfully generic error message (which I neglected to copy, sorry). Before I changed them, they were completing in an hour.

Given my poor SQL skills, perhaps there are straightforward improvements that can be made to the queries I produced, which do the job. But I fear the sheer weight of data we have in the activity event tables may call for a more fundamental change.

Incidentally, I also tried to run an `EXPLAIN` on the updated query, but I canceled that when it still hadn't completed after running for 4 hours. Then I tried an `EXPLAIN` on one of the sub-queries, but it complained  that the table needed to be [analyzed](http://docs.aws.amazon.com/redshift/latest/dg/r_ANALYZE.html). So I kicked off an `ANALYZE daily_multi_device_users`, which is still running as I write this, 4 hours later.

In the meantime, I've been doing some reading around the topic of redshift performance and it seems, to my naive understanding, like we have two options if there are no big wins in the query itself:

1. Reduce the amount of data we have. Either by sampling the events that are emitted from the code, or by expiring old data and employing some kind round-robin approach that aggregates it and allows analysis at lower granularity. Neither approach appeals to me at the moment.

2. No idea how reasonable this is in terms of resource usage, but could we have extra copies of the tables that use a different `DISTKEY`? Currently, we use `uid` as `DISTKEY`, what if we created extra tables that used `service` and `ua_os`? Then the segmented queries could target those tables specifically and, presumably, run much faster.

And of course, it's likely there are other things we can do that I don't know about.

@rfk, care to chip in?

philbooth:
For reference, I am pasting the shorter of the two slow queries (engagement ratio), in it's current state. Would be great if somebody could show me how to reduce the number of subqueries and joins I'm using.

``` sql
WITH
   days AS (
    SELECT DISTINCT day
    FROM daily_multi_device_users
    WHERE day - '28 days'::INTERVAL >= (
      SELECT MIN(day)
      FROM daily_multi_device_users
    )
  ),
  dau AS (
[...]

PR #23, opened 2016-11-07
feat(scripts): vacuum tables after import

philbooth:
Related to https://github.com/mozilla/fxa-activity-metrics/issues/20#issuecomment-258583986, adds `VACUUM` commands for each table after data is added to them.

Strictly speaking, in most cases we only need `SORT ONLY` rather than `FULL`. However the scripts do `DELETE` rows if `force_reload` is set, and the redshift docs state that the saving for `SORT ONLY` isn't normally huge, so it seems prudent to opt for `FULL`.

I timed some `VACUUM` commands earlier and they took about five seconds to complete. I didn't check the stats beforehand though to see whether they had much to do, so that may not be a fair indication of how the import scripts will be affected in practice.

You'll notice the weird extra `END;` that each `VACUUM` block starts with. This is the only way I found to work around the error `VACUUM cannot run inside a transaction block`. I also tried `SET AUTOCOMMIT`, which didn't work.

@rfk, r?

### DIFF
--- a/calculate_daily_summary.py
+++ b/calculate_daily_summary.py
@@ -105,6 +105,12 @@ Q_GET_LAST_AVAILABLE_DAY = """
     FROM activity_events;
 """
 
+Q_VACUUM_TABLES = """
+    END;
+    VACUUM FULL daily_activity_per_device;
+    VACUUM FULL daily_multi_device_users;
+"""
+
 def summarize_events(day_from=None, day_until=None):
     db = postgres.Postgres(DB)
     db.run("BEGIN TRANSACTION")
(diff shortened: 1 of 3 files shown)


rfk:
I'm +1 to this, and I think the vacuum should complete in a timely manner if we do it regularly.

> This is the only way I found to work around the error VACUUM cannot run inside a transaction block

This flips my weirdness sensor though, I'm going to take a look and see if I can understand what's going on here before merge.

rfk:
OK, looks like the "postgres" wrapper module runs everything in a transaction by default, and there's not an easy way to disable that.  We might try moving to the lower-level "psycopg2" module but r+ on this for now.

PR #26, opened 2016-11-08
fix(scripts): increase metrics context property length

### LINKED ISSUE #24: More incidences of columns being too short to store flow data
Following on from #21, I encountered a couple more cases where the length of `VARCHAR` data caused imports to fail:

Field|Value
-----|-----
`entrypoint`|`preferencespilotfiber.texts.at.worktracker.com`
`utm_content`|`fx-reset-passworpourqoi les poils blancs sont plus durs que les noirsd`

Let's update the column definitions here to handle existing data, implement truncation in the content and auth servers when events are emitted, and maybe also consider whether we actually need all of these columns in redshift anyway. Do we really expect to plot data from the `utm_term` and `utm_content` columns?

philbooth:
Fixes #24, restricting all metrics context properties to `VARCHAR(100)`.

Although it's only the `entrypoint` and `utm_content` columns that are currently causing problems for us, a uniform length makes implementation of truncation for mozilla/fxa-auth-server#1539 and mozilla/fxa-content-server#4378 far simpler.

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -48,15 +48,15 @@ Q_CREATE_CSV_TABLE = """
       ua_browser VARCHAR(40),
       ua_version VARCHAR(40),
       ua_os VARCHAR(40),
-      context VARCHAR(40),
-      entrypoint VARCHAR(40),
-      migration VARCHAR(40),
-      service VARCHAR(40),
-      utm_campaign VARCHAR(40),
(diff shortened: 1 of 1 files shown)


philbooth:
Closing this because I just discovered the `TRUNCATECOLUMNS` argument to `COPY`. Going to play around with it to make sure it works how I hope it does, then open a PR for it if so.

PR #31, opened 2016-11-10
feat(summaries): Maintain 28-day rollup summaries of daily activity.

rfk:
Putting this up for discussion only - I haven't tried running it and it's certainly not ready for merge.

As noted in https://github.com/mozilla/fxa-activity-metrics/issues/20#issuecomment-259004981, we commonly need to calculate "monthly X" summaries and it involves a pretty expensive join.  This change would maintain those as a rollup table that we can query directly.

@philbooth thoughts?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,3 +22,4 @@ import: | $(ENV)/COMPLETE
 .PHONY: summarize
 summarize: | $(ENV)/COMPLETE
 	$(ENV)/bin/python ./calculate_daily_summary.py
+	$(ENV)/bin/python ./calculate_monthly_summary.py
--- a/calculate_monthly_summary.py
+++ b/calculate_monthly_summary.py
@@ -1,0 +1,184 @@
+#
+# Script to calculate monthly summary tables from raw activity-event data.
+# It mintain a monthly rollup summarizing users who met the criteria for the
+# `daily_activity_per_device` and `daily_multi_device_users` tables on any
+# day in the preceeding 28 days.
+#
+
+import json
+import time
+import datetime
+
+import postgres
+
+# Load config from disk,
+# and pull in credentials from the environment.
+
+with open("config.json") as f:
+    CONFIG = json.loads(f.read())
+
+DB = "postgresql://{db_username}:{db_password}@{db_host}:{db_port}/{db_name}".format(**CONFIG)
+
+# For the monthly device activity summary,
+# we maintain a table giving unique (uid, device_id, service)
+# values seen in the last 28 days
+
+Q_MONTHLY_ACTIVITY_CREATE_TABLE = """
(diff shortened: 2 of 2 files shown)


calculate_daily_summary.py
```diff
@@ -55,6 +60,48 @@
...
+      ORDER BY 1
+    ) AS days,
+    INNER JOIN daily_activity_per_device AS past
+      ON past.day <= days.day
+      AND past.day > (days.day - '28 days'::INTERVAL)
+      -- This is not strictly necessary,
+      -- but might help speed the query up.
+      AND past.day > ('{day_from}'::DATE - '28 days'::INTERVAL)
+    ORDER BY 1;
+"""
```
rfk: This is subtle, and I'm not sure the naming here is very effective at conveying what it means, but I haven't yet come up with something better.

The table is named "monthly_unique_activity", but it contains data on a daily basis.  There's a record in this table for ($DAY, ....blah....) IFF there's a record in the daily-activity table for any of the 28 days preceeding $DAY.  I guess we could call it "daily_summary_of_unique_activity_over_previous_28_days" but that seems a little unwieldy :-)

Actually "unique_activity_in_previous_month" is not _too_ bad, come to think of it...

rfk:
One downside of this approach: it will use a lot of disk space.  The `daily_activity_per_device` table is already pretty big, and this will be like 28 times the size.

rfk:
I tried out the `monthly_unique_activity` summarization on just the data from November, so around 8 days worth.  It took a whopping 2 hours to complete, and used a lot of temporary disk space, although the final disk usage is modest.  If we do want to do this, we'll need to backfill the calculation a few days at a time.

It was useful though.  Doing a daily MAU query on those 8 days using our current join-at-query-time approach took 6 minutes.  Doing it using the summary table took 45 seconds.  And they reported identical counts for each day.

I'm going to push my luck and try to summarize October as well...

rfk:
> I'm going to push my luck and try to summarize October as well...

LOL, nope, ran out of disk space :-P

philbooth:
@rfk, just got round to properly looking at this now, sorry for the delay. For the record I'm a fan of the approach. Does the disk space issue put a kibosh on it though, or should some combination of reducing the job size and/or increasing the disk size let it finish?

calculate_daily_summary.py
philbooth: For the record, I am a fan of `unique_activity_in_previous_month`. `monthly_unique_activity` does seem like a potential source of minor confusion in the future.

rfk:
I updated this to use a separate script, making it easier to experiment with.  It loads the data in small chunks to avoid using too much temporary disk space.  But I fear the steady-state disk space requirements are still going to be too much to make this worthwhile in practice.

rfk:
Redshift is working great lately, I'm going to close this out.

PR #34, opened 2016-11-24
fix(db): sanely delete rows to avoid duplicate event data

### LINKED ISSUE #33: Flow event CSVs contain overlapping data
On re-running the flow event import just now, I noticed that from the 17th November onwards we receive overlapping timestamps between our daily CSVs:

```
LOADING 81 DAYS OF DATA
LOADING 2016-11-23
  MIN TIMESTAMP 1474316017
  MAX TIMESTAMP 1479968087
LOADING 2016-11-22
  MIN TIMESTAMP 1474616313
  MAX TIMESTAMP 1479879483
LOADING 2016-11-21
  MIN TIMESTAMP 1470904755
  MAX TIMESTAMP 1479821935
LOADING 2016-11-20
  MIN TIMESTAMP 1476423151
  MAX TIMESTAMP 1479697472
LOADING 2016-11-19
  MIN TIMESTAMP 1474316017
  MAX TIMESTAMP 1479626260
LOADING 2016-11-18
  MIN TIMESTAMP 1476519693
  MAX TIMESTAMP 1479526463
LOADING 2016-11-17
  MIN TIMESTAMP 1470239187
  MAX TIMESTAMP 1479491045
LOADING 2016-11-16
  MIN TIMESTAMP 1479254400
  MAX TIMESTAMP 1479340799
```

Currently the import scripts assume that there is no overlap and try to avoid duplicates by first deleting all data for the day that is being imported. It looks like we can't rely on that assumption any more. Presumably this is a side effect of mozilla-services/puppet-config#2368.

I've so far failed to imagine a robust solution to this in the code, but maybe we could add a column to `flow_events` that identifies the originating CSV, then use that column for deletion/avoiding duplicates instead.

philbooth:
Fixes #33.

I haven't tested this out at all yet, so there may be syntax errors, logic errors and lord knows what else. But I wanted to propose something concrete-ish as an approach to fix the duplicate events problem.

@rfk, what do you think?

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -89,23 +89,31 @@ Q_CREATE_EVENTS_TABLE = """
       -- but redshift doesn't support that.
       flow_time BIGINT NOT NULL ENCODE lzo,
       flow_id VARCHAR(64) NOT NULL DISTKEY ENCODE lzo,
-      type VARCHAR(64) NOT NULL ENCODE lzo
+      type VARCHAR(64) NOT NULL ENCODE lzo,
+      export_date DATE NOT NULL ENCODE lzo
     );
 """
 
 Q_CHECK_FOR_DAY = """
     SELECT timestamp FROM flow_events
-    WHERE timestamp::DATE = '{day}'::DATE
+    WHERE timestamp::DATE <= '{day}'::DATE + '1 day'::INTERVAL
+      AND timestamp::DATE >= '{day}'::DATE - '1 day'::INTERVAL
+      AND export_date = '{day}'::DATE;
     LIMIT 1;
 """
 
-Q_CLEAR_DAY_METADATA = """
-    DELETE FROM flow_metadata
-    WHERE begin_time::DATE = '{day}'::DATE;
-"""
 Q_CLEAR_DAY_EVENTS = """
     DELETE FROM flow_events
-    WHERE timestamp::DATE = '{day}'::DATE;
+    WHERE timestamp::DATE <= '{day}'::DATE + '1 day'::INTERVAL
+      AND timestamp::DATE >= '{day}'::DATE - '1 day'::INTERVAL
+      AND export_date = '{day}'::DATE;
+"""
+Q_CLEAR_DAY_METADATA = """
+    DELETE flow_metadata
+    FROM flow_metadata AS m
+    LEFT JOIN flow_events AS e
+      ON m.flow_id = e.flow_id
+    WHERE e.flow_id IS NULL;
 """
 
 Q_COPY_CSV = """
@@ -165,8 +173,13 @@ Q_INSERT_METADATA = """
       utm_medium,
       utm_source,
       utm_term
-    FROM temporary_raw_flow_data
-    WHERE type LIKE 'flow%begin';
+    FROM temporary_raw_flow_data AS d
+    WHERE type LIKE 'flow%begin'
+      AND NOT EXISTS (
+        SELECT 1
+        FROM flow_metadata AS m
+        WHERE m.flow_id = d.flow_id
+      );
 """
 Q_UPDATE_DURATION = """
     UPDATE flow_metadata
@@ -234,13 +247,15 @@ Q_INSERT_EVENTS = """
       timestamp,
       flow_time,
       flow_id,
-      type
+      type,
+      export_date
     )
     SELECT
       'epoch'::TIMESTAMP + timestamp * '1 second'::INTERVAL,
       flow_time,
       flow_id,
-      type
+      type,
+      '{day}'::DATE
     FROM temporary_raw_flow_data;
 """
 
@@ -279,8 +294,8 @@ def import_events(force_reload=False):
             # Create the temporary table
             db.run(Q_CREATE_CSV_TABLE)
             # Clear any existing data for the day, to avoid duplicates.
-            db.run(Q_CLEAR_DAY_METADATA.format(day=day))
             db.run(Q_CLEAR_DAY_EVENTS.format(day=day))
+            db.run(Q_CLEAR_DAY_METADATA)
             s3path = EVENTS_FILE_URL.format(day=day)
             # Copy data from s3 into redshift
             db.run(Q_COPY_CSV.format(
@@ -294,7 +309,7 @@ def import_events(force_reload=False):
             db.run(Q_UPDATE_NEW_ACCOUNT)
             db.run(Q_UPDATE_METRICS_CONTEXT)
             # Populate the flow_events table
-            db.run(Q_INSERT_EVENTS)
+            db.run(Q_INSERT_EVENTS.format(day=day))
             # Print the timestamps for sanity-checking.
             print "  MIN TIMESTAMP", db.one("SELECT MIN(timestamp) FROM temporary_raw_flow_data")
             print "  MAX TIMESTAMP", db.one("SELECT MAX(timestamp) FROM temporary_raw_flow_data")


import_flow_events.py
```diff
@@ -100,12 +101,15 @@
 """
 
 Q_CLEAR_DAY_METADATA = """
-    DELETE FROM flow_metadata
-    WHERE begin_time::DATE = '{day}'::DATE;
+    DELETE m
+    FROM flow_metadata AS m
+    LEFT JOIN flow_events AS e
+      ON m.flow_id = e.flow_id
+    WHERE e.flow_id IS NULL;
```
rfk, line 116: I have a hard time understanding this query - is it "delete flow_metadata records that do not have any events"?

This raises the question of, how well do we cope with flows whose events are spread across the date boundary between two files?  I can imagine situations where we're deleting and re-import some-but-not-all of the events in a flow.

import_flow_events.py
```diff
@@ -100,12 +101,15 @@
...
 """
 Q_CLEAR_DAY_EVENTS = """
     DELETE FROM flow_events
-    WHERE timestamp::DATE = '{day}'::DATE;
+    WHERE origin = '{day}'::DATE;
```
rfk: One trick here is that we're no longer deleting by the sortkey, so this query will have to examine every item in the table, and hence will get more and more expensive over time.  Perhaps we could add additional `WHERE` clauses to restrict `timestamp` to be within 1 day either side of the target, so that it can avoid scanning rows that are definitely not going to match.

import_flow_events.py
```diff
@@ -89,7 +89,8 @@
       -- but redshift doesn't support that.
       flow_time BIGINT NOT NULL ENCODE lzo,
       flow_id VARCHAR(64) NOT NULL DISTKEY ENCODE lzo,
-      type VARCHAR(64) NOT NULL ENCODE lzo
+      type VARCHAR(64) NOT NULL ENCODE lzo,
+      origin DATE NOT NULL ENCODE lzo
```
rfk: Nit: it's not obvious what the name "origin" means without looking at this script; something like "importDate" or "sourceFileDate" or similar might make it clearer.

rfk:
@philbooth I think you're onto something here, the approach makes sense to me.  I haven't thought through all the implications of handling flows that cross a file boundary though.

import_flow_events.py
rfk: Ah OK, I think I see - we delete the event data first, and then we delete any metadata for flows that not longer have any events.  I think this will work well for the case where you're re-importing the later half of a flow, but I'm not sure about when you're importing the earlier half (i.e. you're deleting and re-importing the flow.begin event, but not re-importing the events that complete it).  But it's not obvious what should happen in that case anyway.

philbooth: > ...this will work well for the case where you're re-importing the later half of a flow, but I'm not sure about when you're importing the earlier half (i.e. you're deleting and re-importing the flow.begin event, but not re-importing the events that complete it).

I've been thinking about this overnight and the only sane approach I can think of involves special-casing the `INSERT` to `flow_metadata`, so that it checks for the existence of that `flow_id` before inserting. I'm going to play around with this approach a bit now, then I'll come back and update the PR (or not).

import_flow_events.py
philbooth: > ...it's not obvious what the name "origin" means without looking at this script...

Good point.

> ...something like "importDate" or "sourceFileDate" or similar might make it clearer.

Fwiw, I think `export_date` feels more logical than `import_date` to me. Any good?

(maybe it's just a reflection of how much I've been re-importing this data recently, but the import time for those CSVs feels completely independent to when they were dumped in S3 to me)

philbooth:
This approach (with some fixes made since I opened the PR) seems to work okay in terms of importing the data. It's slower than it was (obviously), but I don't see a good way to avoid that and still be tolerant of the overlaps.

philbooth:
> It's slower than it was...

That slowness gets noticeably worse as the import proceeds through more iterations. It will be interesting to see how an import of a single day works out, after the main import is finished. I'm hopeful that, post-`VACUUM`, it won't be too bad.

philbooth:
> It's slower than it was...

Just to be clear, as I realise I didn't say this out loud before, the slow(er) bit is `Q_CLEAR_DAY_METADATA`.

philbooth:
> It's slower than it was...

I significantly underestimated the order of magnitude here. It's still running now, ~10 hours later. The job has reached `2016-09-15`, still two weeks shy of day zero. Exponential growth? 😱

philbooth:
> It's still running now...

Going to leave it running although I'm aware things may go haywire if the cron-scheduled job kicks in and this one has not completed. But, weekend, let's see what happens. If anybody wants to cancel this job, it is pid 7798.

import_flow_events.py
rfk: SGTM

import_flow_events.py
```diff
@@ -89,23 +89,31 @@
...
+    WHERE timestamp::DATE <= '{day}'::DATE + '1 day'::INTERVAL
+      AND timestamp::DATE >= '{day}'::DATE - '1 day'::INTERVAL
+      AND export_date = '{day}'::DATE;
+"""
+Q_CLEAR_DAY_METADATA = """
+    DELETE flow_metadata
+    FROM flow_metadata AS m
+    LEFT JOIN flow_events AS e
+      ON m.flow_id = e.flow_id
+    WHERE e.flow_id IS NULL;
```
rfk, line 116: This is almost certainly quadratic in the size of the tables, as it will have to compare every row of `flow_metadata` with every row of `flow_events`.  You may be able to speed it up by restricting the timestamps, along the lines of:

```
m.begin_time::DATE >= {day}::DATE - '1 day'::INTERVAL AND m.begin_time::DATE <= {day}::DATE
```

To only scan part of the flow_metadata table, and:

```
e.timestamp >= m.begin_time AND e.timestamp::DATE < m.begin_time::DATE + '1 day'::INTERVAL
```

To only scan part of the `flow_events` table.  Not sure how much difference it will make in practice though.

philbooth: > Not sure how much difference it will make in practice though.

Maybe I did it wrong but this actually made it even slower when I tried it just now.

I've reinstated the script from `master` and kicked off a fresh import. Just fyi in case you happen to be looking at flow data during the course of your Monday.

rfk: Interesting.  I wonder if the cost of looking up that extra column for each row, outweighs the intended benefit of reducing the data to be scanned.

rfk: Also it occurs to me that you don't have to limit the event timestamp by `m.begin_time`, you can just use `{day}` directly:

```
e.timestamp >= >= {day}::DATE - '1 day'::INTERVAL AND e.timestamp::DATE <= {day}::DATE + '1 day'::INTERVAL
```

Which might make it perform better overall.

vladikoff:
from mtg: @rfk to chat with @philbooth about this....

import_flow_events.py
rfk: @philbooth I wonder if we really need the join here, or if we can get away with something like:
 
* Add an `export_date` column to the flow_metadata table; it will be the export date of the flow.begin event.
* When clearing a day, delete from flow_metadata where `export_date` matches the date in question

What bad things might happen if we just treat the row in `flow_metadata` as belonging to the `flow.begin` event like this?

philbooth: > What bad things might happen if we just treat the row in `flow_metadata` as belonging to the `flow.begin` event like this?

Would that lead to (a small number of) duplicated events at the boundaries? Or am I not thinking it through properly? Either way, I'll give it a bash at the weekend.

I also had a plan to try out adding a separate table with `export_date DISTKEY` and `flow_id SORTKEY` in the hope that would be quicker, but maybe that's wishful thinking.

I'm opting for the weekend so it matters less if I break stuff for a bit.

rfk: > Would that lead to (a small number of) duplicated events at the boundaries?

Maybe.  If you've got a concrete example scenario in mind it would be great to write it out explicitly.

I see two possibilities - the events from the start of a flow being deleted and re-imported, and events from the end of a flow being deleted and re-imported.

If we're deleting events from the end of the flow, then we are not deleting the `flow.begin` event and so the flow will just stick around in the `flow_metadata` table.  We might re-calculate its various properties like "completed" and "new_account", but we won't add a duplicate entry for it.

If we're deleting events from the start of the flow, then we delete the `flow.begin` event and hence delete the row from `flow_metadata`.  It will be re-created when we re-import the events for that day.  But we have to be careful to ensure that it gets the appropriate values for "completed", "new_account" etc, because those events might have come from the following day, and are not being re-imported.

philbooth: > If you've got a concrete example scenario in mind it would be great to write it out explicitly.

Nope, your exposition above makes sense. Reviewing your other PR now, I'll close this one out.

philbooth:
Closing this in favour of #39.

PR #37, opened 2016-12-13
fix(flow-events): Use timestamp filters on flow metadata update queries.

rfk:
This helps them scan less of the flow_metadata table and hence to run faster - on a test of importing the latest 5 days of events, it dropped the load time from around 30mins to around 20mins.

@philbooth r?


philbooth (approved):
r+, nice one.

And if we decide to drop the older flow events at any point, we can reduce the spread further by losing the `+`.

Issue #19, opened 2016-11-02
Remove ${viewName} from begin events, move it to flow_metadata


philbooth:
We should probably add a `begin_view` column to the `flow_metadata` table and populate it based on the `viewName` part of `flow.${viewName}.begin`. We could then set the `type` of those begin events to just `flow.begin`, which might make consuming our event data easier for people who lack all the context of content server implementation details. It also makes `flow.begin` nicely symmetrical with `flow.complete`.

Stupidly, when I first thought about this I was envisaging a knock-on change to the content server  and the Heka filter, to set the view name as a field in the event data. But of course there's no need for that, we can just do a separate `UPDATE` after the `INSERT` and pull it out of the `type`, similar to what we do for `duration` etc.

rfk:
We could even consider an explicit event for "saw a view", so the flow would become:
- flow.begin
- flow.${viewName}.view
- flow.${viewName}.engage
- ...
- flow.complete

The timing data on the `flow.${viewName}.view` event would be particularly interesting, as it would more-or-less correspond to the network transit time for our client-side assets.

vladikoff:
cc @philbooth do you want this in active or next 2 weeks?

philbooth:
Yep, next two weeks, 100%, final answer.

philbooth:
This was fixed by #44.

Issue #43, opened 2016-12-16
Dont compress our sortkeys


rfk:
There's an interesting note on [1] that says:

"""
We do not recommend applying runlength encoding on any column that is designated as a sort key. Range-restricted scans perform better when blocks contain similar numbers of rows. If sort key columns are compressed much more highly than other columns in the same query, range-restricted scans might perform poorly.
"""

We're not using run-length encoding, but we *are* compressing our sortkey.  The presentation at [2] contains this advice:

"""
If your sort keys compress significantly more than your data columns, you may want to skip compression of sortkey column(s).

Check SVV_TABLE_INFO(skew_sortkey1)
"""

And indeed, SVV_TABLE_INFO tells me that `skew_sortkey` for our main tables is on the order of several hundred, which seems high.  We might find there's a performance win if we disable compression of our sortkeys.  It's probably worth an experiment.

[1] http://docs.aws.amazon.com/redshift/latest/dg/c_Runlength_encoding.html
[2] http://www.slideshare.net/AmazonWebServices/bdt401-amazon-redshift-deep-dive-tuning-and-best-practices

philbooth:
I've finally managed to measure this change against a subset of our activity event data. The bottom line is that I detected no noticeable improvement.

I created two subsets of the activity event data for the period 1st November until 17th December, one with compressed SORTKEYs and the other uncompressed. These are available in the following tables:

* `daily_multi_device_users_test_compressed`
* `daily_activity_per_device_test_compressed`
* `daily_multi_device_users_test_uncompressed`
* `daily_activity_per_device_test_uncompressed`

I then ran modified versions of the engagement ratio / multi-device query, against each pair of tables. Those queries are available here:

* https://sql.telemetry.mozilla.org/queries/1942/source#3426
* https://sql.telemetry.mozilla.org/queries/1943/source#3436

The timings showed a lot of variation when other queries were running but converge around the 8-minute mark in both cases:

* Compressed: 8'06", 8'33", 13'21", 8'34", 7'27"
* Uncompressed: 8'03", 7'41", 23'02", 7'50", 7'58"

Perhaps with a bigger sample it would be possible to detect a more reliable difference, but it took over 2 days to set up the sample data for this small test so I'm not sure it's worth the effort.

[...]

PR #44, opened 2016-12-17
Performance and correctness fixes for the flow import scripts

### LINKED ISSUE #33: Flow event CSVs contain overlapping data

### LINKED ISSUE #40: flow_metadata table should use flow_id as distkey
For consistency with the flow_events table.  Currently it does not have a distkey.

philbooth:
This changeset contains three significant commits:

* `flow_metadata.flow_id` is promoted to `DISTKEY`, to match `flow_events.flow_id`. This fixes #40.

* Compression is disabled on `flow_metadata.begin_time` and `flow_events.timestamp`. This is a partial fix for #43.

* Sane deletion logic for overlapping flow data is implemented via `flow_metadata.export_date` and `flow_events.export_date` columns. This a combination of #39 and #34, with a couple of fixes for stupid syntax errors (my fault). Fixes #33.

For expediency, or perhaps it was just laziness, I tested the effects of these changes all together. I hope that's okay, because they can still be measured independently to some extent.

The import time should only be affected by the `export_date` commit. Performance of the analytical queries is affected by both the `DISTKEY` and the `SORTKEY` commits, but it seemed like the latter was the only one we were unsure about there so I opted to measure them together. And we can get independent validation of the `SORTKEY` change by further measuring its impact against a subset of the activity event data, which I plan to do tomorrow.

So the results, based on 104 days of data (2nd September until 16th December), are like so:

**Complete import duration:** ~9.5 hours
**Of which, vacuuming duration:** ~45 minutes

**Independently clearing the 104th day:**  ~20 seconds
**Independently importing the 104th day:** ~7.5 minutes

**Executing the engagement ratio/multi-device query:** ~1.5 hours

The import time is a huge improvement on my original ham-fisted attempt to fix duplicate events in #34. It's a little bit worse than the last timing I have for importing without the fix, which was ~6 hours for 84 days of data. The single-day clear/import figures seem eminently reasonable too.

The performance of the engagement ratio query is a significant improvement over what we currently have. My last timing was ~2 hours when run against 96 days of data.

In terms of correctness, everything seems to look okay. I can conceive of one theoretical problem but it shouldn't harm us in practice. Before the content server implemented flow event timestamp validation (train 74), we emitted events that were more than one day removed from their `export_date`. In such cases, were we to clear and then import a single day, it would lead to duplicating those events with the current logic.

However, in practice, I don't expect us to re-import any of that historical data on a per-day basis and the  described problem does not occur when re-importing wholesale. Furthermore, given that we know the data prior to train 74 contains garbage, I fully expect us to permanently delete it all once we've built up a bit more history to show longer trends in the charts.

All told, I believe these changes are a big improvement on what we currently have. @rfk, r?

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -79,7 +79,8 @@ Q_CREATE_METADATA_TABLE = """
       utm_content VARCHAR(40) ENCODE lzo,
       utm_medium VARCHAR(40) ENCODE lzo,
       utm_source VARCHAR(40) ENCODE lzo,
-      utm_term VARCHAR(40) ENCODE lzo
+      utm_term VARCHAR(40) ENCODE lzo,
+      export_date DATE NOT NULL ENCODE lzo
     );
 """
 Q_CREATE_EVENTS_TABLE = """
@@ -89,23 +90,30 @@ Q_CREATE_EVENTS_TABLE = """
       -- but redshift doesn't support that.
       flow_time BIGINT NOT NULL ENCODE lzo,
       flow_id VARCHAR(64) NOT NULL DISTKEY ENCODE lzo,
-      type VARCHAR(64) NOT NULL ENCODE lzo
+      type VARCHAR(64) NOT NULL ENCODE lzo,
+      export_date DATE NOT NULL ENCODE lzo
     );
 """
 
 Q_CHECK_FOR_DAY = """
     SELECT timestamp FROM flow_events
-    WHERE timestamp::DATE = '{day}'::DATE
+    WHERE timestamp::DATE >= '{day}'::DATE - '1 day'::INTERVAL
+      AND timestamp::DATE <= '{day}'::DATE + '1 day'::INTERVAL
+      AND export_date = '{day}'::DATE
     LIMIT 1;
 """
 
-Q_CLEAR_DAY_METADATA = """
-    DELETE FROM flow_metadata
-    WHERE begin_time::DATE = '{day}'::DATE;
-"""
 Q_CLEAR_DAY_EVENTS = """
     DELETE FROM flow_events
-    WHERE timestamp::DATE = '{day}'::DATE;
+    WHERE timestamp::DATE >= '{day}'::DATE - '1 day'::INTERVAL
+      AND timestamp::DATE <= '{day}'::DATE + '1 day'::INTERVAL
+      AND export_date = '{day}'::DATE;
+"""
+Q_CLEAR_DAY_METADATA = """
+    DELETE FROM flow_metadata
+    WHERE begin_time::DATE >= '{day}'::DATE - '1 day'::INTERVAL
+      AND begin_time::DATE <= '{day}'::DATE + '1 day'::INTERVAL
+      AND export_date = '{day}'::DATE;
 """
 
 Q_COPY_CSV = """
@@ -148,7 +156,8 @@ Q_INSERT_METADATA = """
       utm_content,
       utm_medium,
       utm_source,
-      utm_term
+      utm_term,
+      export_date
     )
     SELECT
       flow_id,
@@ -164,7 +173,8 @@ Q_INSERT_METADATA = """
       utm_content,
       utm_medium,
       utm_source,
-      utm_term
+      utm_term,
+      '{day}'::DATE
     FROM temporary_raw_flow_data
     WHERE type LIKE 'flow%begin';
 """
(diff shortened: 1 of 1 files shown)


rfk:
> Before the content server implemented flow event timestamp validation (train 74), 
> we emitted events that were more than one day removed from their export_date.

Would it be worthwhile to add logic that drops such events before import, e.g. by deleting anything in `temporary_raw_flow_data` that has a timestamp outside the affected range?

> The performance of the engagement ratio query is a significant improvement over
> what we currently have

So I gotta be honest...I don't understand how the changes to the flow-event tables can have affected this query over activity-events.  Unless I'm misunderstanding what you're saying above.

import_flow_events.py
```diff
@@ -61,8 +61,8 @@
 """
 Q_CREATE_METADATA_TABLE = """
     CREATE TABLE IF NOT EXISTS flow_metadata (
-      flow_id VARCHAR(64) NOT NULL UNIQUE ENCODE lzo,
```
rfk, line 64: Do we lose anything by dropping the `UNIQUE` here?  Redshift doesn't enforce the uniqueness, but it does allegedly use it to speed up queries in some circumstances.  (Indeed, I wonder if `flow_id` should be marked as `PRIMARY KEY` in this table).

rfk (approved):
One question re: the unique constraint, but otherwise this LGTM.  Nice work @philbooth, r+

philbooth:
> Would it be worthwhile to add logic that drops such events before import, e.g. by deleting anything in `temporary_raw_flow_data` that has a timestamp outside the affected range?

Possibly, yep. Or it may be simpler to just run a separate clean-up script once against the affected CSVs and scrub the future events. I don't know why I didn't do that already tbh.

> I don't understand how the changes to the flow-event tables can have affected this query over activity-events.

Oh dear, that is an excellent point. I'm talking complete gibberish, sorry. For some reason, I was completely mixed up about what tables each query runs against, I don't think my brain is working properly post-Hawaii yet.

Annoyingly, I now don't have anything to make the before-and-after comparison against because I didn't record any numbers for the time-to-device-connection query.

I'm still working on measuring the `SORTKEY` change against a subset of the activity event data though, so hopefully it's okay to extrapolate from there?

rfk:
> I'm still working on measuring the SORTKEY change against a subset of the activity
> event data though, so hopefully it's okay to extrapolate from there?

👍

import_flow_events.py
philbooth: > Do we lose anything by dropping the `UNIQUE` here?

I don't know for sure. Based on reading the docs beforehand I didn't think so but...

> Redshift doesn't enforce the uniqueness, but it does allegedly use it to speed up queries in some circumstances.

I didn't realise this. Where did you read about the performance part? I confess I am on shaky ground with much of the key-related stuff in redshift, some of the docs could do with more concrete detail imho.

rfk: See e.g. http://docs.aws.amazon.com/redshift/latest/dg/t_Defining_constraints.html although that only talks about primary keys, not general unique constraints.

philbooth:
Changes pushed:

* `UNIQUE` constraint reinstated as per https://github.com/mozilla/fxa-activity-metrics/pull/44#discussion_r92954918. I couldn't find anything to suggest that it has any benefit, but I also couldn't find anything to suggest it does any harm.
* `SORTKEY` compression has been reinstated, see discussion in #43.
* Old-style `flow.${viewName}.begin` events are mapped to `flow.begin` on import, just to make querying against old data more consistent in light of mozilla/fxa-content-server#4440. Fixes #19.

I've also added a `WIP` label to the PR because I plan to try these changes out in anger tonight, just to make sure I haven't broken anything.

philbooth:
Removed the `WIP` from this and carrying forward @rfk's earlier r+. These changes are running well against our redshift instance and the import cron jobs have been reinstated.

Issue #45, opened 2016-12-20
Implement automated expiry of old data


philbooth:
We've tried a few different tricks to improve query performance but maybe the slowness is a basic facet of the number of events in our tables. This is especially true of activity events, where we have a historical record going back to the start of 2016.

To address this, we could apply a sample rate to the events once they reach an expiry age. This could be handled fairly crudely by a script running, say, every Saturday and deleting *n* of every *nn* events older than a certain date (ordered by timestamp). We could also make this expiry gradual if we're concerned about abrupt boundaries affecting analysis.

For instance:

* age < 3 months, no events deleted; 100% sample rate.
* age >= 3 months && age < 6 months, every other event (when ordered by timestamp) is deleted; 50% sample rate.
* age >= 6 months && age < 2 years, 9 of every 10 events (when ordered by timestamp) are deleted; 10% sample rate.
* age >= 2 years, all events deleted; 0% sample rate.

The reason I suggest ordering by timestamp and then deleting at regular intervals is to preserve the natural daily/seasonal ebb and flow in event frequency. That should hopefully ensure the general shape of our charts remains consistent as expiration progresses.

@rfk, thoughts?

philbooth:
I'm not sure I'd thought this through very clearly when I wrote it down. The most obvious problem with it is that writing queries against a variable sample rate sounds like a complete nightmare. Perhaps a `sample_rate` or `multiplier` column could be added to help with that but it's still pretty awkward.

The other thing that just occurred to me, is the expiration query itself would need to be very careful not to re-sample already-sampled data. The aforementioned extra column would help with that too I guess, but it still feels like it violates the principle of least astonishment somewhat.

rfk:
Some related thoughts in https://github.com/mozilla/fxa-activity-metrics/issues/42

I agree that having different sample-rates of data in the same table is probably asking for trouble.  But I can imagine us maintaining e.g.:

* `activity_events` at 100% sample rate, but only the most recent 3 months of data
* `activity_events_20pct` at 20% sample rate, but only the most recent 6 months of data
* `activity_events_10pct` at 10% sample rate, with a full 2 years of data

That's a little bit of duplication, but probably not very much in the scheme of things, if we choose the sample rates and time intervals properly.

philbooth:
Ah, yeah, separate tables sounds much cleaner, you're right.

And it seems like this is a total dupe of #42, sorry. I'll close this one.

PR #46, opened 2016-12-22
wip: proposal for sampled activity event imports

### LINKED ISSUE #42: Maintain down-sampled tables for faster ad-hoc analysis?
Our tables are very very big.  It turns out (surprise!) that this can really slow down queries if you have to use a nested join.

It might be worthwhile us maintaining an `activity_events_sampled_10pc` table alongside the `activity_events` table, containing data for only 10% of uids.  This would be more performant for exploratory analysis, but still a big enough sample to be meaningful.  (And it's what we used to have when we first started out with these metrics).

Of course, the downside would be the overhead of maintaining a second set of tables, both in terms of disk usage, and in terms of extra scripting etc.  But it may be worth a little experiment.

philbooth:
Fixes #42.

@rfk, this is another one of my poorly-thought-through hacks, opened with a view to getting your feedback if you have time to look at it quickly. As usual, I haven't tried running it yet so it there's probably logic and syntax errors.

It represents my first stab at what the import script for activity events might look like, with sampling of old data implemented as per #42 / #45.

~~Hopefully it's quite straightforward, the only bit of magic I used was the [`ROW_NUMBER` window function](http://docs.aws.amazon.com/redshift/latest/dg/r_WF_ROW_NUMBER.html) to sample the data on insert into each table.~~

In its current state it would need to be run as a full, clean re-import the first time round, or perhaps it could be preceded by a dedicated one-off script that sets up the sampled tables using the already-imported dataset.

~~And obviously there would need to be matching changes to the `calculate_daily_summary` script too.~~

Thoughts?

### DIFF
--- a/import_activity_events.py
+++ b/import_activity_events.py
@@ -41,9 +41,9 @@ EVENTS_FILE_FIXED_URL = "s3://" + EVENTS_BUCKET + "/" + EVENTS_PREFIX_FIXED + "e
 EVENTS_BEGIN = datetime.strptime("2016-09-02", "%Y-%m-%d")
 
 SAMPLE_RATES = (
-    {"percent":10, "months":24, "table_suffix":"_sampled_10"},
-    {"percent":50, "months":6, "table_suffix":"_sampled_50"},
-    {"percent":100, "months":3, "table_suffix":""}
+    {"percent":10, "months":24, "suffix":"_sampled_10"},
+    {"percent":50, "months":6, "suffix":"_sampled_50"},
+    {"percent":100, "months":3, "suffix":""}
 )
 
 # We import each into a temporary table and then
@@ -65,7 +65,7 @@ Q_CREATE_CSV_TABLE = """
 """
 
 Q_CREATE_EVENTS_TABLE = """
-    CREATE TABLE IF NOT EXISTS activity_events{table_suffix} (
+    CREATE TABLE IF NOT EXISTS activity_events{suffix} (
       timestamp TIMESTAMP NOT NULL SORTKEY ENCODE lzo,
       uid VARCHAR(64) NOT NULL DISTKEY ENCODE lzo,
       type VARCHAR(30) NOT NULL ENCODE lzo,
@@ -84,7 +84,7 @@ Q_CHECK_FOR_DAY = """
 """
 
 Q_CLEAR_DAY = """
-    DELETE FROM activity_events{table_suffix}
+    DELETE FROM activity_events{suffix}
     WHERE timestamp::DATE = '{day}'::DATE;
 """
 
@@ -106,7 +106,7 @@ Q_COPY_CSV = """
 """
 
 Q_INSERT_EVENTS = """
-    INSERT INTO activity_events{table_suffix} (
+    INSERT INTO activity_events{suffix} (
       timestamp,
       uid,
       type,
@@ -117,7 +117,7 @@ Q_INSERT_EVENTS = """
       ua_os
     )
     SELECT
-      'epoch'::TIMESTAMP + timestamp * '1 second'::INTERVAL,
+      ts,
       uid,
       type,
       device_id,
@@ -126,20 +126,24 @@ Q_INSERT_EVENTS = """
       ua_version,
       ua_os
     FROM (
-      SELECT *, STRTOL(SUBSTRING(uid FROM 0 FOR 8), 16) % 100 AS cohort
+      SELECT
+        *,
+        'epoch'::TIMESTAMP + timestamp * '1 second'::INTERVAL AS ts,
+        STRTOL(SUBSTRING(uid FROM 0 FOR 8), 16) % 100 AS cohort
       FROM temporary_raw_activity_data
     )
-    WHERE cohort <= {sample_rate};
+    WHERE cohort <= {percent}
+      AND ts::DATE <= '{day}'::DATE + '{months} months'::INTERVAL;
 """
 
 Q_DELETE_EVENTS = """
-    DELETE FROM activity_events{table_suffix}
-    WHERE timestamp::DATE >= '{date}'::DATE + '{months} months'::INTERVAL;
+    DELETE FROM activity_events{suffix}
+    WHERE timestamp::DATE > '{day}'::DATE + '{months} months'::INTERVAL;
 """
 
 Q_VACUUM_TABLES = """
     END;
-    VACUUM FULL activity_events{table_suffix};
+    VACUUM FULL activity_events{suffix};
 """
 
 def import_events(force_reload=False):
@@ -147,7 +151,7 @@ def import_events(force_reload=False):
     db = postgres.Postgres(DB)
     db.run(Q_DROP_CSV_TABLE)
     for rate in SAMPLE_RATES:
-        db.run(Q_CREATE_EVENTS_TABLE.format(table_suffix=rate["table_suffix"]))
+        db.run(Q_CREATE_EVENTS_TABLE.format(suffix=rate["suffix"]))
     days = []
     days_to_load = []
     # Find all the days available for loading.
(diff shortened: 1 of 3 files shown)


import_activity_events.py
```diff
@@ -119,19 +125,29 @@
       ua_browser,
       ua_version,
       ua_os
-    FROM temporary_raw_activity_data;
+    FROM (
+      SELECT *, ROW_NUMBER() OVER(ORDER BY timestamp, uid, type) AS row
+      FROM temporary_raw_activity_data
+    )
+    WHERE row % {modulo} = 0;
```
rfk: I think doing this by row number will make it hard to analyse the data, because you'll have some-but-not-all of the events from any particular user.  That makes it hard to do e.g. meaningful retention analysis, since we wouldn't know whether the user stopped using firefox, or just happend to be outside the sample set.

It may be better to sample 10% of userids rather than 10% of events, like we used to do in the S3 export.  We could use something like this to convert the uid to an integer for sampling"

```
 SELECT uid, STRTOL(SUBSTRING(uid FROM 0 FOR 8), 16) % 100 AS sampleNum FROM activity_events LIMIT 10;
                               uid                                | samplenum 
------------------------------------------------------------------+-----------
 b0ae7947a2aff25fdd19f55eb948a1eecda728c4c0e6eb3538c715deb5e5be8d |        20
 7653f1590e47dbbeaa9110c979b7159c361b69c4cef04a8cfa029a3ede93d9c6 |        97
 a4f7f93ea4c9edace3b630c51c3e999fb11fa041036035c030589fd002ac0034 |        63
 c8ab8bf2685a29c72d3a8d6f1eba7388573cb813b1701cac688f5bae24ace47f |        55
 58960ea4307b7738694a554ea76110f04e1735160b858a29c9ab64e9e0cb6364 |        22
 5d904e947dee5c84226cb7d76695c0edb688b13fdd94404200b3f1aa219da784 |        49
 d487c8ff86788a3a02eeb0f2d859f1f914ba40dc6534d9c3bc484760f16942e6 |        87
 705f2f5098db22514cdf097845675755f0f9e035f7fdf0da497be9c889517193 |        89
 17947fc6718702699e273d042eeec46c779be4087060a123eea303a996d664ed |         0
 6eb1f20b4f400b13e7196c1ea2e6a8c06962113c8a58c9df3f3cb4a4951031c9 |        24
(10 rows)
```

import_activity_events.py
```diff
@@ -59,7 +65,7 @@
 """
 
 Q_CREATE_EVENTS_TABLE = """
-    CREATE TABLE IF NOT EXISTS activity_events (
+    CREATE TABLE IF NOT EXISTS activity_events_{sample_rate} (
```
rfk: Renaming the main `activity_events` table will break all the existing re:dash graphs.  It's probably much uglier in code, but can we keep `activity_events` for the full export and only add suffixes to the sampled tables?

philbooth: This assumes that the existing queries are only interested in the last 3 months of data (or whatever number we decide on). Is that the case?

In that sense, I kind of saw the enforced name change as a feature. Query authors should make a conscious decision whether they prefer breadth or depth.

(although also, I'm happy to change it if you think that's senseless tyranny)

rfk: Hmm, good point.  TBH I think many of the existing queries may need to move into rollups anyway.  Let's with @davismtl about this before making any final decisions (and feel free to do so without me if the timing works out, since I'm PTO next week)

philbooth:
@rfk, this branch has been updated to incorporate your feedback and also to update the summary tables.

In terms of testing it out, obviously I don't want to mess with the current tables and break everyone's charts. Does it sound reasonable for me to set up a new redshift instance and load the data into sampled tables there? (I've no idea if it was easy or difficult when you set up the current instance)

rfk:
> Does it sound reasonable for me to set up a new redshift instance and load
> the data into sampled tables there?

Sounds reasonable, think it was fairly easy to do through the AWS web console.  A couple of things to keep in mind:

* You'll need to configure access rules to allow access from at least the redshift-helper box; copying the rules from the existing instance is probably fine
* You won't be able to hook it up to re:dash, but that's likely not necessary anyway

philbooth:
> Sounds reasonable, think it was fairly easy to do through the AWS web console.

Thanks, you were right. The import is now running but, given that there is more than a year of data to be imported, I'm not expecting everything to finish for a day or two (just fyi).

vladikoff:
from mtg: need to ref the bug

philbooth:
The down-sampled import is complete! The summary tables aren't there yet, I kicked those off just now.

Here are some numbers.

First up, I timed the import so we know what to expect if we decide to do this for real:

* `import_activity_events_retro.py` took ~25 hours for 245 days of data.
* `import_activity_events.py` took ~13 hours for 130 days of data.

Then after all of that, running `import_activity_events.py` again to import a single day, expire old data and vacuum took ~20 minutes.

I ran some sanity checks to make sure the downsampling appears sane:

```
fxa=# select count(*) as events, count(distinct timestamp::date) as days, min(timestamp::date) as expiry, max(timestamp::date) as now
fxa-# from activity_events;
   events   | days |   expiry   |    now
------------+------+------------+------------
 2016575107 |   93 | 2016-10-10 | 2017-01-10
(1 row)

fxa=# select count(*) as events, count(distinct timestamp::date) as days, min(timestamp::date) as expiry, max(timestamp::date) as now
fxa-# from activity_events_sampled_50;
   events   | days |   expiry   |    now
------------+------+------------+------------
 2010024173 |  185 | 2016-07-10 | 2017-01-10
(1 row)

fxa=# select count(*) as events, count(distinct timestamp::date) as days, min(timestamp::date) as expiry, max(timestamp::date) as now
fxa-# from activity_events_sampled_10;
  events   | days |   expiry   |    now
-----------+------+------------+------------
 841650488 |  376 | 2016-01-01 | 2017-01-10
(1 row)
```

So around 2 billion events per table with this approach as it stands (`_sampled_10` should get to nearly that size with 2 full years of data). Obviously we can control that number by tweaking the expiry, up or down.

I haven't done a perf test using one of our long-running queries yet, because I want to wait for the daily summary tables to finish before I do that. Will post here again after.

philbooth:
`calculate_daily_summary.py` has now finished. It took ~17 hours to run against 376 days of data.

Numbers for `daily_activity_per_device`:

```
fxa=# select count(*) as rows, count(distinct day) as days, min(day) as expiry, max(day) as latest
fxa-# from daily_activity_per_device;
   rows    | days |   expiry   |   latest
-----------+------+------------+------------
 678511752 |   93 | 2016-10-10 | 2017-01-10
(1 row)

fxa=# select count(*) as rows, count(distinct day) as days, min(day) as expiry, max(day) as latest
fxa-# from daily_activity_per_device_sampled_50;
   rows    | days |   expiry   |   latest
-----------+------+------------+------------
 679684460 |  185 | 2016-07-10 | 2017-01-10
(1 row)

fxa=# select count(*) as rows, count(distinct day) as days, min(day) as expiry, max(day) as latest
fxa-# from daily_activity_per_device_sampled_10;
   rows    | days |   expiry   |   latest
-----------+------+------------+------------
 239558039 |  309 | 2016-03-08 | 2017-01-10
(1 row)
```

Numbers for `daily_multi_device_users`:

```
fxa=# select count(*) as rows, count(distinct day) as days, min(day) as expiry, max(day) as latest
fxa-# from daily_multi_device_users;
   rows    | days |   expiry   |   latest
-----------+------+------------+------------
 396364040 |   93 | 2016-10-10 | 2017-01-10
(1 row)

fxa=# select count(*) as rows, count(distinct day) as days, min(day) as expiry, max(day) as latest
fxa-# from daily_multi_device_users_sampled_50;
   rows    | days |   expiry   |   latest
-----------+------+------------+------------
 388266447 |  185 | 2016-07-10 | 2017-01-10
(1 row)

fxa=# select count(*) as rows, count(distinct day) as days, min(day) as expiry, max(day) as latest
fxa-# from daily_multi_device_users_sampled_10;
   rows    | days |   expiry   |   latest
-----------+------+------------+------------
 170943672 |  308 | 2016-03-09 | 2017-01-10
(1 row)
```

Performance test results to follow.

philbooth:
Oh, just spotted the `_sampled_10` summary tables have less data than I was expecting. Looking at the query now to try and understand why.

philbooth:
> the `_sampled_10` summary tables have less data than I was expecting.

The same is true of the (unsampled) summary tables in the `rfkelly-fxa-test` cluster. Initially I wondered if that was when `device_id` got added to the activity event data, because those queries rely on `device_id`. But looking at the git logs, that doesn't seem to be it.

[This change](https://github.com/mozilla/fxa-auth-server/pull/1213/files) to the activity events happened not long after the drop-off, but I can't see how it's related.

@rfk, any ideas? Am I missing something obvious here?

philbooth:
Timings for the engagement ratio / multi-device query against each table set:

* unsampled: 22 minutes
* 50% sampled: 27 minutes
* 10% sampled: 25 minutes

rfk:
> Am I missing something obvious here?

I don't see anything obviously wrong with the queries; can you say more about the oddness you're seeing and why it doesn't match your expectations?

philbooth:
> ...can you say more about the oddness you're seeing and why it doesn't match your expectations?

It's nothing to worry about fwiw, I figured it out since then. But anyway, since you ask, the thing I did not understand was:

Why does the data in our daily summary tables only go back to March?

We have data from 2016-01-01, but `daily_activity_per_device` begins on 2016-03-08 and `daily_multi_device_users` begins on 2016-03-09.

The answer is the `device_id`, even though it didn't look like the answer when I checked the git log:

```
fxa=# select min(timestamp) from activity_events where device_id != '';
         min
---------------------
 2016-03-08 23:17:34
(1 row)
```

So as expected, in other words.

philbooth:
@rfk, if you're interested in playing with the down-sampled tables, you can connect to redshift with `~/pb-downsampled/connect.sh` on the `fxa-redshift-helper` instance.

philbooth:
Closing this in favour of #47, because reasons.

PR #47, opened 2017-01-24
feat(import): maintain down-sampled tables with automatic data expiry

### LINKED ISSUE #42: Maintain down-sampled tables for faster ad-hoc analysis?

philbooth:
Fixes #42.

This is the non-wip version of #46. I opted to do it as a clean PR and summarise the lengthy discussion from over there as a single, succinct-ish description.

Queries against the activity event data are slow, because we have so many events. That is a problem which will get worse over time. This change attempts to fix it by reducing the size of the data.

It does so by maintaining three sets of activity event data:

* Unsampled, holding 3 months of data.
* 50% sampled, holding 6 months of data.
* 10% sampled, holding 2 years of data.

The values for sample rate and history length are essentially arbitrary, although I picked those in an effort to have roughly the same number of records in each set. In theory, most queries should take about the same amount of time regardless of the set they're running against.

In practice, I ran a couple of queries against them and got these results:

Query|Data set|Execution time
-----|--------|--------------
[Retention](https://sql.telemetry.mozilla.org/queries/725/source)|100% / 3 months|1 hour, 14 minutes
[Retention](https://sql.telemetry.mozilla.org/queries/725/source)|50% / 6 months|57 minutes
[Retention](https://sql.telemetry.mozilla.org/queries/725/source)|10% / 13 months*|13 minutes
[Engagement](https://sql.telemetry.mozilla.org/queries/2333/source)|100% / 3 months|25 minutes
[Engagement](https://sql.telemetry.mozilla.org/queries/2333/source)|50% / 6 months|32 minutes
[Engagement](https://sql.telemetry.mozilla.org/queries/2333/source)|10% / 13 months*|18 minutes

(*only 13 months because we don't have enough history for the full 2 years yet)

For comparison, the execution times for those queries against our existing data set are:

Query|Data set|Execution time
-----|--------|--------------
[Retention](https://sql.telemetry.mozilla.org/queries/725/source)|100% / 13 months|~10 hours (estimated)
[Engagement](https://sql.telemetry.mozilla.org/queries/2333/source)|100% / 13 months|2 hours, 20 minutes

Tweaking the values for sample rate and history length up or down, to tailor performance to our needs, is a straightforward change to the scripts.

It may be that we decide to maintain these tables alongside the existing ones, rather than in replacement of. If that is the case, we could just rename these altered queries, or perhaps we'd want them in a separate repo.

@rfk, r?

### DIFF
--- a/import_activity_events.py
+++ b/import_activity_events.py
@@ -40,6 +40,15 @@ EVENTS_FILE_URL = "s3://" + EVENTS_BUCKET + "/" + EVENTS_PREFIX + "events-{day}.
 EVENTS_FILE_FIXED_URL = "s3://" + EVENTS_BUCKET + "/" + EVENTS_PREFIX_FIXED + "events-{day}.fixed.csv"
 EVENTS_BEGIN = datetime.strptime("2016-09-02", "%Y-%m-%d")
 
+# We have sampled data sets that include a longer history.
+# This config controls the tables names, sample rates and
+# history length for each data set.
+SAMPLE_RATES = (
+    {"percent":10, "months":24, "suffix":"_sampled_10"},
+    {"percent":50, "months":6, "suffix":"_sampled_50"},
+    {"percent":100, "months":3, "suffix":""}
+)
+
 # We import each into a temporary table and then
 # INSERT them into the activity_events table
 
@@ -59,7 +68,7 @@ Q_CREATE_CSV_TABLE = """
 """
 
 Q_CREATE_EVENTS_TABLE = """
-    CREATE TABLE IF NOT EXISTS activity_events (
+    CREATE TABLE IF NOT EXISTS activity_events{suffix} (
       timestamp TIMESTAMP NOT NULL SORTKEY ENCODE lzo,
       uid VARCHAR(64) NOT NULL DISTKEY ENCODE lzo,
       type VARCHAR(30) NOT NULL ENCODE lzo,
@@ -72,13 +81,13 @@ Q_CREATE_EVENTS_TABLE = """
 """
 
 Q_CHECK_FOR_DAY = """
-    SELECT timestamp FROM activity_events
+    SELECT timestamp FROM activity_events_sampled_10
     WHERE timestamp::DATE = '{day}'::DATE
     LIMIT 1;
 """
 
 Q_CLEAR_DAY = """
-    DELETE FROM activity_events
+    DELETE FROM activity_events{suffix}
     WHERE timestamp::DATE = '{day}'::DATE;
 """
 
@@ -100,7 +109,7 @@ Q_COPY_CSV = """
 """
 
 Q_INSERT_EVENTS = """
-    INSERT INTO activity_events (
+    INSERT INTO activity_events{suffix} (
       timestamp,
       uid,
       type,
@@ -111,7 +120,7 @@ Q_INSERT_EVENTS = """
       ua_os
     )
     SELECT
-      'epoch'::TIMESTAMP + timestamp * '1 second'::INTERVAL,
+      ts,
       uid,
       type,
       device_id,
@@ -119,20 +128,33 @@ Q_INSERT_EVENTS = """
       ua_browser,
       ua_version,
       ua_os
-    FROM temporary_raw_activity_data;
+    FROM (
+      SELECT
+        *,
+        'epoch'::TIMESTAMP + timestamp * '1 second'::INTERVAL AS ts,
+        STRTOL(SUBSTRING(uid FROM 0 FOR 8), 16) % 100 AS cohort
+      FROM temporary_raw_activity_data
+    )
+    WHERE cohort <= {percent}
+      AND ts::DATE >= '{day}'::DATE - '{months} months'::INTERVAL;
+"""
+
+Q_DELETE_EVENTS = """
+    DELETE FROM activity_events{suffix}
+    WHERE timestamp::DATE < '{day}'::DATE - '{months} months'::INTERVAL;
 """
 
 Q_VACUUM_TABLES = """
     END;
-    VACUUM FULL activity_events;
+    VACUUM FULL activity_events{suffix};
 """
 
 def import_events(force_reload=False):
     b = boto.s3.connect_to_region("us-east-1").get_bucket(EVENTS_BUCKET)
     db = postgres.Postgres(DB)
     db.run(Q_DROP_CSV_TABLE)
-    db.run(Q_CREATE_EVENTS_TABLE)
-    days = []
+    for rate in SAMPLE_RATES:
+        db.run(Q_CREATE_EVENTS_TABLE.format(suffix=rate["suffix"]))
     days_to_load = []
     # Find all the days available for loading.
     for key in b.list(prefix=EVENTS_PREFIX):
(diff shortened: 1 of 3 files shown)


calculate_daily_summary.py
```diff
@@ -90,62 +101,73 @@
...
-        db.run(Q_DAILY_DEVICES_CREATE_TABLE)
-        db.run(Q_MD_USERS_CREATE_TABLE)
-        # By default, summarize the latest days that are not yet summarized.
-        if day_from is None:
-            day_from = db.one(Q_GET_FIRST_UNPROCESSED_DAY)
+        for suffix in TABLE_SUFFIXES:
+            db.run(Q_DAILY_DEVICES_CREATE_TABLE.format(suffix=suffix))
+            db.run(Q_MD_USERS_CREATE_TABLE.format(suffix=suffix))
+            day_expiry = db.one(Q_GET_FIRST_AVAILABLE_DAY.format(suffix=suffix))
+            # By default, summarize the latest days that are not yet summarized.
```
rfk: nit: this "by default..." doesn't really apply anymore, because you can't pass in the dates as arguments, right?

calculate_daily_summary.py
```diff
@@ -90,62 +101,73 @@
...
     try:
-        db.run(Q_DAILY_DEVICES_CREATE_TABLE)
-        db.run(Q_MD_USERS_CREATE_TABLE)
-        # By default, summarize the latest days that are not yet summarized.
-        if day_from is None:
-            day_from = db.one(Q_GET_FIRST_UNPROCESSED_DAY)
+        for suffix in TABLE_SUFFIXES:
+            db.run(Q_DAILY_DEVICES_CREATE_TABLE.format(suffix=suffix))
+            db.run(Q_MD_USERS_CREATE_TABLE.format(suffix=suffix))
+            day_expiry = db.one(Q_GET_FIRST_AVAILABLE_DAY.format(suffix=suffix))
```
rfk: nit: it took me a while to work out how `day_expiry` worked here.  IIUC, since it's the first day of data available in the main events table, we delete any items older than that from the summary table.  I wonder if we can rename it something a little clearer like `day_earliest` or `day_$BIKESHEDCOLOR`, but not a big deal.

calculate_daily_summary.py
```diff
@@ -90,62 +101,73 @@
...
-        print "UPDATING MULTI-DEVICE USERS SUMMARY"
-        db.run(Q_MD_USERS_CLEAR.format(**days))
-        db.run(Q_MD_USERS_SUMMARIZE.format(**days))
+            day_until = db.one(Q_GET_LAST_AVAILABLE_DAY.format(suffix=suffix))
+            days = {
+                "day_from": day_from,
+                "day_until": day_until,
+                "suffix": suffix
+            }
+            print "SUMMARIZING FROM", day_from, "UNTIL", day_until
```
rfk: nit: maybe worth adding "FOR SUFFIX" to these prints, like you did with the expiry one below.

import_activity_events.py
```diff
@@ -40,6 +40,12 @@
 EVENTS_FILE_FIXED_URL = "s3://" + EVENTS_BUCKET + "/" + EVENTS_PREFIX_FIXED + "events-{day}.fixed.csv"
 EVENTS_BEGIN = datetime.strptime("2016-09-02", "%Y-%m-%d")
 
+SAMPLE_RATES = (
+    {"percent":10, "months":24, "suffix":"_sampled_10"},
+    {"percent":50, "months":6, "suffix":"_sampled_50"},
+    {"percent":100, "months":3, "suffix":""}
+)
```
rfk, line 50: This might deserve a couple of lines commend explaining what it's for and how it maps to the tables in the db.

import_activity_events.py
```diff
@@ -111,27 +117,41 @@
...
+        STRTOL(SUBSTRING(uid FROM 0 FOR 8), 16) % 100 AS cohort
+      FROM temporary_raw_activity_data
+    )
+    WHERE cohort <= {percent}
+      AND ts::DATE >= '{day}'::DATE - '{months} months'::INTERVAL;
+"""
+
+Q_DELETE_EVENTS = """
+    DELETE FROM activity_events{suffix}
+    WHERE timestamp::DATE < '{day}'::DATE - '{months} months'::INTERVAL;
```
rfk, line 144: You could do a sub-select here like `timestamp::DATE < (SELECT MAX(timestamp::DATE) FROM activity_events{suffix}) - '{months} months'::INTERVAL` if you didn't want to have to pas in `{day}` as a parameter.  It's not obvious to me how strictly we're guaranteed that the day passed in here is the largest day in the table.

rfk (approved):
@philbooth I am impressed how cleanly this came together, nice work!  I left a handful of nits but r+ with or without any changes, at your discretion.

I'd like to try to get this deployed into the existing redshift instance, in a way that's minimally-invasive to the existing tables.  Will it work if we just run it as-is, truncating the existing tables to their new expiry date and creating the new ones from scratch?

Or, would we have to move the "days_to_load" logic into the `for rate in SAMPLE_RATES` loop for that to work?

import_activity_events.py
philbooth: > It's not obvious to me how strictly we're guaranteed that the day passed in here is the largest day in the table.

I thought we'd guaranteed it pretty solidly, but I may be guilty of wooly thinking. `days_to_load` is sorted is sorted into reverse-order and the dates are in format `YYYY-MM-DD`. Having said that, I appreciate the benefit of putting that logic in the same place as the query itself, rather than depending on context from elsewhere in the script. Will change.

philbooth:
> Will it work if we just run it as-is, truncating the existing tables to their new expiry date and creating the new ones from scratch?

It *should* just work with the tables in their current state, because the `Q_CHECK_FOR_DAY` query runs against the 10% sampled table, which will be empty at that point. However I haven't actually tested that out for real, to be sure. I've only run them against an empty database and then iteratively against their own generated state.

So possibly I'd be more comfortable dropping the tables first and running it from scratch against an empty db. That whole operation, start to finish, can fit into a weekend to minimise disruption.

The other option would be to have a separate, one-off script that duplicates the current tables with the `_sampled_nn` suffixes and then just does the expiring parts of these scripts standalone. Then these scripts would be left to take over in cron. But I'm not sure if we have enough disk space configured to cope with three entire copies of the data, so this might not be an option.

philbooth:
> The other option would be to have a separate, one-off script that duplicates the current tables with the `_sampled_nn` suffixes and then just does the expiring parts of these scripts standalone.

*Duh*, of course, it could just copy them using an appropriate `WHERE` clause and avoid the need for separate expiry queries entirely.

vladikoff:
from mtg: @rfk and @philbooth to discuss this...

philbooth:
These completed successfully earlier, so I'm going to carry the forward the r+ from three weeks ago and finally merge this.

PR #53, opened 2017-02-22
fix(scripts): ensure dependent flow_metadata columns are set sanely

### LINKED ISSUE #51: Flows that contain `flow.complete`, but have `completed=false`
Here's something odd I noticed:

  https://sql.telemetry.mozilla.org/queries/3026/source?p_interval=7

That query selects flows that contain a `flow.complete` event, but for which the `flow_metadata.completed` column has not been set to true.  It looks like they're all in the most recently-imported day.  @philbooth any clue why this might be happening?  Will they get correctly marked as `completed=true` when the next daily import runs?

philbooth:
Fixes #51.

This needs to be accompanied with an equivalent update to the existing `flow_metadata` to properly fix the issue. That could be done with a bespoke standalone query or we could just drop the data and reimport it using these changes over the weekend. The flow event import has less moving parts than the activity event import, so I expect it be a smoother ride than last weekend if we take that route.

@rfk r?

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -178,6 +178,24 @@ Q_INSERT_METADATA = """
     FROM temporary_raw_flow_data
     WHERE type LIKE 'flow%begin';
 """
+
+Q_INSERT_EVENTS = """
+    INSERT INTO flow_events (
+      timestamp,
+      flow_time,
+      flow_id,
+      type,
+      export_date
+    )
+    SELECT
+      'epoch'::TIMESTAMP + timestamp * '1 second'::INTERVAL,
+      flow_time,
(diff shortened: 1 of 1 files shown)


import_flow_events.py
```diff
@@ -253,23 +271,6 @@
       AND flow_metadata.begin_time <= '{day}'::DATE + '1 day'::INTERVAL;
 """
 
-Q_INSERT_EVENTS = """
```
philbooth, line 256: I like to keep these definitions in invocation order to make reading them more logical, since they're a bit distant from the point of usage.

rfk (approved):
LGTM

PR #54, opened 2017-02-22
feat(scripts): add optional locale and uid to flow_metadata

### LINKED ISSUE #36: Add optional uid field to flow_metadata
I cleared this with the Data Steward in at the all-hands.  Let's add an additional "uid" column to the flow_metadata table, and record the uid involved for flows that get far enough along for us to know it.

In conjunction with https://github.com/mozilla/fxa-content-server/issues/4290, this will help us to a/b test improvements to our signup/signin process.

### LINKED ISSUE #48: Capture Language (or locale) in Meta Data
Copied from @davismtl's report in https://github.com/mozilla/fxa-content-server/issues/4694

Context:
During a recent firstrun a/b test, we observed that our control was performing much better than the page in production. Normally these pages are identical. The exception in the A/B test is that the control was only displayed to EN-US.

This means that there are other languages, locales or geographic regions that are really under performing.

We need to save language or locale.

Related: https://waffle.io/mozilla/fxa/cards/57bb62ccf7a3eaa80292e2f8 (add country)

philbooth:
Fixes #36. Fixes #48.

This is rebased against #53 because merge conflicts. You might want to review that one before this one, or both together is fine too.

mozilla-services/puppet-config#2488 was merged last night, which means the CSV files going forward will contain `locale` and `uid` fields. This change updates the import scripts so that we populate them correctly.

We could add the new columns to the existing `flow_metadata` schema with a separate query but, given that #53 needs doing at the same time, I'm minded to drop all the data and re-import it. Two birds, one stone, zero bespoke queries.

@rfk r?

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -56,7 +56,9 @@ Q_CREATE_CSV_TABLE = """
       utm_content VARCHAR(40),
       utm_medium VARCHAR(40),
       utm_source VARCHAR(40),
-      utm_term VARCHAR(40)
+      utm_term VARCHAR(40),
+      locale VARCHAR(40),
+      uid VARCHAR(64)
     );
 """
 Q_CREATE_METADATA_TABLE = """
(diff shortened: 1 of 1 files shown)


philbooth:
Oh, just realised I haven't tackled the problem of handling older, pre-locale/uid CSVs here yet. Closing.

PR #55, opened 2017-02-22
feat(scripts): add optional locale and uid to flow_metadata

### LINKED ISSUE #36: Add optional uid field to flow_metadata

### LINKED ISSUE #48: Capture Language (or locale) in Meta Data

philbooth:
Fixes #36. Fixes #48. Replaces #54.

This is rebased against #53 because merge conflicts. You might want to review that one before this one, or both together is fine too.

mozilla-services/puppet-config#2488 was merged last night, which means the CSV files going forward will contain `locale` and `uid` fields. This change updates the import scripts so that we populate new columns for these fields when they're present.

We could add the new columns to the existing `flow_metadata` schema with a separate query but, given that #53 needs doing at the same time, I'm minded to drop all the data and re-import it. Two birds, one stone, zero bespoke queries.

@rfk r?

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -178,6 +178,24 @@ Q_INSERT_METADATA = """
     FROM temporary_raw_flow_data
     WHERE type LIKE 'flow%begin';
 """
+
+Q_INSERT_EVENTS = """
+    INSERT INTO flow_events (
+      timestamp,
+      flow_time,
+      flow_id,
+      type,
+      export_date
+    )
+    SELECT
+      'epoch'::TIMESTAMP + timestamp * '1 second'::INTERVAL,
+      flow_time,
+      flow_id,
+      (CASE WHEN type LIKE 'flow.%.begin' THEN 'flow.begin' ELSE type END),
+      '{day}'::DATE
+    FROM temporary_raw_flow_data;
+"""
+
 Q_UPDATE_DURATION = """
     UPDATE flow_metadata
     SET duration = durations.flow_time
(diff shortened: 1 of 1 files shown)


import_flow_events.py
```diff
@@ -289,7 +301,6 @@ def import_events(force_reload=False):
     for key in b.list(prefix=EVENTS_PREFIX):
         filename = os.path.basename(key.name)
         day = "-".join(filename[:-4].split("-")[1:])
-        print day
```
philbooth, line 292: This little guy outlived his usefulness a long time ago and just wastes precious newlines in my terminal buffer now.

import_flow_events.py
```diff
@@ -32,6 +32,13 @@
 EVENTS_PREFIX = "fxa-flow/data/"
 EVENTS_FILE_URL = "s3://" + EVENTS_BUCKET + "/" + EVENTS_PREFIX + "flow-{day}.csv"
 
+# The CSV format has changed over time. This config controls the
+# column names, definitions and deploy dates for those changes.
+ADDITIONAL_COLUMNS = (
+    {"name":"locale", "definition":"VARCHAR(40)", "deploy_date":"2017-02-21"},
+    {"name":"uid", "definition":"VARCHAR(64)", "deploy_date":"2017-02-21"},
```
philbooth: Fwiw, it turns out this deploy date is not correct. I assumed it would be from the merge, but the CSV for `2017-02-22` still contains 15 commas per line. I'll keep an eye on the CSVs then come back to update this as and when.

philbooth: As per https://bugzilla.mozilla.org/show_bug.cgi?id=1339405#c1, the extra fields begin part-way through the 23rd. I updated the dates here to match.

rfk (approved):
This LGTM; it's  bit yuck to manage the additional columns in code but I like that it will get us up and running without messing with the historical data :-)

philbooth:
It turns out I had unjustified optimism about how well the `COPY` query would handle mixed-length CSV rows. So to make this work I had to pre-process `2017-02-23` CSV in awk to make this work correctly. With that change, plus the usual dosage of trial and error to fix the obligatory SQL syntax errors, this is now running for real. Merging.

PR #58, opened 2017-03-09
feat(scripts): add scripts for cleaning and padding the flow data

philbooth:
@rfk, these are the housekeeping scripts I mentioned in IRC earlier/yesterday. Feel free to merge if you think they'll add value, or ignore if you think they're redundant now that the validation is all sorted.

### DIFF
--- a/clean-flow-data.sh
+++ b/clean-flow-data.sh
@@ -1,0 +1,47 @@
+#!/bin/sh
+
+if [ "$#" != 1 ]; then
+  echo "Usage: $0 <DAY>"
+  exit 1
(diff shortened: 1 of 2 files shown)


philbooth:
(actually the cleaned data is still not importing correctly yet, maybe withold your judgement on this one for now)

rfk:
> Feel free to merge if you think they'll add value,

I think they're worth having and they won't be a huge maintenance burden, merge 'em in when you're happy with 'em.

philbooth:
Now we're good.

PR #59, opened 2017-03-10
feat(vendor-export): Prep anonymized data for sending to external vendor for evaluation

rfk:
Sharing this for visibillity, esp. with @davismtl.  This is a script that will take our historical data in S3, and attempt to produce an even-more-anonymized event stream suitable for sharing with external vendors to evaluate their service.  It writes the processed data back to S3 so that we don't have to process it over and over.

### DIFF
--- a/amplitude/backfill_flow_uids.py
+++ b/amplitude/backfill_flow_uids.py
@@ -1,0 +1,100 @@
+
+import os
+import hmac
+import json
+import math
+import time
+import hashlib
+import datetime
+import tempfile
+import urlparse
+
+import postgres
+
+import boto.s3
+import boto.provider
+
+# Load config from disk,
+# and pull in credentials from the environment.
+
+with open("config.json") as f:
+    CONFIG = json.loads(f.read())
+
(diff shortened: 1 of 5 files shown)


davismtl (approved):
AFAICT , everything looks good. I admit that while reading this I was reminded of something I heard during the Amplitude conference this week. "Try to standardize your event names across data tools."

I'll leave that as food for thought for now and we can think about it again and perhaps review things if we proceed past a POC.

philbooth:
> Try to standardize your event names across data tools.

Fwiw, as much as I think this is a noble goal, our flow event names thus far have been a train-wreck of inconsistency (my fault btw). If we can ditch those and impose some order in a shiny new place, I am strongly in favour.

rfk:
FYI I've updated this to do the heavy-lifting in redshift instead of python, since the performance seems comparable and there's less to go wrong.  I'm going to leave it run for a while and see how many days it can churn through...

rfk:
This is running nicely, and I don't think we need to merge it to master since it's just a proof-of-concept integration.  Closing.

PR #62, opened 2017-03-17
refactored import scripts

### LINKED ISSUE #61: Add processing for the experiment flow events
Flow events were added for active experiments in mozilla/fxa-content-server#4775. We should process them in the import scripts so that the experiment data is available as a separate table, joined by `flow_id`.

philbooth:
Fixes #61.

~~Not opening this for comment yet (although comments are still welcome of course), more to raise visibility~~. This is currently deployed to the redshift helper so if the imports suddenly fail there, it's likely because of something I've done wrong here.

~~As I write this, the only change here is to refactor `import_activity_events.py` so that most of the code is pulled out to a generic `import_events.py`. Next step is to use the refactored code to add automatic expiration to the flow import.~~

~~Obviously, everything related to `flow_metadata` will remain specific to the flow import, which may mean looping twice, first through the CSVs and then through the imported days in `flow_events`. Or more likely, I'll pass some kind of command function in to `import_events.py` so it can return control to the calling script on each iteration.~~

Anyway, the point is I'm expecting what's here to evolve rapidly.


import_events.py
```diff
@@ -0,0 +1,235 @@
...
+    DROP TABLE IF EXISTS {table};
+""".format(table=TABLE_NAMES["temp"])
+
+Q_CREATE_EVENTS_TABLE = """
+    CREATE TABLE IF NOT EXISTS {table} (
+        timestamp TIMESTAMP NOT NULL SORTKEY ENCODE lzo,
+        {schema}
+    );
+""".format(table=TABLE_NAMES["perm"],
+           schema="{schema}")
```
philbooth: This partial string-formatting quickly gets out of hand you'll notice. I imagine there must be a cleaner way to do it, feel free to suggest something. Maybe I've gone too far in trying to refactor the table names out of the queries. My aim was to only have to write each table name once, so that if they change they only have to change in one place too.

philbooth:
Taking the `WIP` off this and calling it ready for review.

It does two things:

1. Refactors common code from `import_activity_events.py` and `import_flow_events.py` in to `import_events.py`.

2. Adds a new `flow_experiments` table (and downsampled versions thereof).

(there's also another change to tighten up the cleaning script, but hopefully that's not too contentious)

Here's some sample experiment data:

```
fxa=# select experiment, cohort, "timestamp", substring(flow_id, 0, 8) as flow_id, substring(uid, 0, 8) as uid, export_date
fxa-# from flow_experiments_sampled_10
fxa-# limit 3;
      experiment      |  cohort   |      timestamp      | flow_id |   uid   | export_date
----------------------+-----------+---------------------+---------+---------+-------------
 connectAnotherDevice | treatment | 2017-03-24 23:59:17 | 1438838 |         | 2017-03-25
 connectAnotherDevice | treatment | 2017-03-24 23:58:28 | 68a209a | f87458a | 2017-03-25
 connectAnotherDevice | treatment | 2017-03-24 23:59:06 | 31f273c | f87458a | 2017-03-25
(3 rows)
```

I'm interested in feedback on the new schema, the SQL and on the Python itself.

There's still scope for more refactoring in the future:

* We can eliminate duplicated logic for the `flow_metadata` and `flow_experiments` data sets.

* I'd like to invoke the summary script directly from `import_activity_events.py`, so that we can eliminate the separate `make summarize` step in cron.

I figured it made sense to get the changes here reviewed first before setting off down those roads though.

There were some question marks about import slowness with these changes, which I suspected was caused by excessive load from [insane redash scheduling](https://bugzilla.mozilla.org/show_bug.cgi?id=1323840). In order to prove that, I set up a fresh redshift cluster with the same settings and ran the full import there. Within 10 hours, it had overtaken the import job on our current cluster, which had been running for 4 days at that point. It finished the entire import in less than two days (not sure of the exact figure because I didn't `time` it and it finished while I was asleep). As things stand, the main import still has about 90 days left to import and we are approaching the 5-day mark.

Sadly, the redash scheduler is still going nuts, rendering our main cluster mostly useless at the moment:

Screenshot showing duplicate queries spawned by the redash scheduler

I wonder, would it be okay to keep my "other" redshift cluster running to act as a kind of fast replica for dev/debugging work? Waiting for stuff to finish on the main cluster is pretty frustrating at the moment. It's called `fxa-pb-proof`, in case anyone wants to play with it (but let's not schedule anything from redash to run there 😄).

import_flow_events.py
```diff
@@ -93,69 +86,25 @@
       uid VARCHAR(64) ENCODE lzo
     );
 """
-Q_CREATE_EVENTS_TABLE = """
-    CREATE TABLE IF NOT EXISTS flow_events (
+
+Q_CREATE_EXPERIMENTS_TABLE = """
+    CREATE TABLE IF NOT EXISTS flow_experiments{suffix} (
+      experiment VARCHAR(40) NOT NULL DISTKEY ENCODE lzo,
+      cohort VARCHAR(40) NOT NULL ENCODE lzo,
```
philbooth, line 93: Initially I tried to call this `group` to match the nomenclature from the content server. Then I realised that clashes with the SQL keyword so decided to opt for something else. I find the need to constantly quote or namespace the `timestamp` column a bit annoying, and didn't want to add more of the same.

philbooth:
Oh, I forgot to mention the third thing it does! It also down-samples the flow data, whoops! 😊

import_events.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def get_timestamp(which):
+        return db.one(Q_GET_TIMESTAMP.format(which=which, event_type=event_type))
+
+    def print_timestamp(which):
+        print "  {which} timestamp".format(which=which), get_timestamp(which)
+
+    def import_day(day):
+        print day
+        print "  COPYING CSV"
```
philbooth, line 173: I've been doing a *lot* of debugging of these scripts over the past couple of weeks and found the tweaks to the log messages helpful when scanning back through the output to see what's happened. Essentially I aimed for:

* Ouputting the minimal amount of information per line, to make it easy to parse.
* Indenting lines to emphasise hierarchy and thus reduce any need for repetition.

import_events.py
```diff
@@ -0,0 +1,236 @@
...
+
+Q_DELETE_EVENTS = """
+    DELETE FROM {table}
+    WHERE timestamp::DATE < '{day}'::DATE - '{months} months'::INTERVAL;
+""".format(table=TABLE_NAMES["perm"], day="{day}", months="{months}")
+
+Q_VACUUM_TABLES = """
+    END;
+    VACUUM FULL {table};
+    ANALYZE {table};
```
philbooth, line 120: Fwiw, we weren't doing this `ANALYZE` step previously, so weren't getting all of the optimization benefit available to us.

vladikoff:
from mtg: @rfk to take a look at this

import_events.py
rfk: oh, interesting, nice catch

import_flow_events.py
```diff
@@ -278,88 +220,137 @@
...
+      'epoch'::TIMESTAMP + timestamp * '1 second'::INTERVAL,
+      flow_id,
+      uid,
+      '{day}'::DATE
+    FROM (
+      SELECT *, STRTOL(SUBSTRING(flow_id FROM 0 FOR 8), 16) % 100 AS sample
+      FROM {table_name}
+    )
+    WHERE sample <= {percent}
+    AND type LIKE 'flow.experiment%';
```
rfk: nit: should this be `flow.experiment.%` so that it doesn't match e.g. `flow.experimentalSomethingOrOther` events we might add in future?

rfk (approved):
This all looks OK to me, definitely a cleanup from the previous heavy duplication.

One style thing to consider: the calls to `import_events.run()` might read a little clearer if you use explicit names for the args, like:

```
import_events.run(
    s3_prefix="foo/bar",
    event_type="activity",
    temp_schema=SCHEMA,
    ...and so on...
)
```

PR #67, opened 2017-03-29
feat(scripts): Add script to import email events

vbudhram:
This is still WIP script to import email events into redash.

### DIFF
--- a/import_email_events.py
+++ b/import_email_events.py
@@ -1,0 +1,18 @@
+#
+# Script to import "email event" metrics from S3 into redshift.
+#
+
+import import_events
+SCHEMA = """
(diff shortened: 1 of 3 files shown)


import_email_events.py
```diff
@@ -0,0 +1,18 @@
+#
+# Script to import "email event" metrics from S3 into redshift.
+#
+
+import import_events
+SCHEMA = """
+    timestamp BIGINT NOT NULL SORTKEY,
```
philbooth: I realise I haven't documented this very well (read: at all), but the idea was that event tables will always have a `timestamp` column so you don't need to explicitly define it here. `import_events` will always define it on your behalf.

PR #70, opened 2017-04-12
feat(scripts): Add email events import script

vbudhram:
This PR adds the script to import email events into redshift.

### DIFF
--- a/import_email_events.py
+++ b/import_email_events.py
@@ -1,0 +1,24 @@
+#
+# Script to import "email event" metrics from S3 into redshift.
+#
+
+import import_events
+SCHEMA = """
+    flow_id VARCHAR(64) DISTKEY ENCODE lzo,
+    domain VARCHAR(40) ENCODE lzo,
+    template VARCHAR(64) ENCODE lzo,
+    type VARCHAR(64) NOT NULL ENCODE lzo,
+    complaint VARCHAR(64) ENCODE lzo,
+    bounced VARCHAR(64) ENCODE lzo,
+    locale VARCHAR(64) ENCODE lzo
+"""
+
+COLUMNS = "flow_id, domain, template, type, complaint, bounced, locale"
+
+import_events.run(s3_prefix="fxa-email/data/email-events",
+                  event_type="email",
+                  temp_schema=SCHEMA,
+                  temp_columns=COLUMNS,
+                  perm_schema=SCHEMA,
+                  perm_columns=COLUMNS,
+                  id_column="flow_id")
--- a/import_email_events.py
+++ b/import_email_events.py
@@ -8,12 +8,12 @@ SCHEMA = """
     domain VARCHAR(40) ENCODE lzo,
     template VARCHAR(64) ENCODE lzo,
     type VARCHAR(64) NOT NULL ENCODE lzo,
-    complaint VARCHAR(64) ENCODE lzo,
     bounced VARCHAR(64) ENCODE lzo,
+    complaint VARCHAR(64) ENCODE lzo,
     locale VARCHAR(64) ENCODE lzo
 """
 
-COLUMNS = "flow_id, domain, template, type, complaint, bounced, locale"
+COLUMNS = "flow_id, domain, template, type, bounced, complaint, locale"
 
 import_events.run(s3_prefix="fxa-email/data/email-events",
                   event_type="email",


import_email_events.py
```diff
@@ -0,0 +1,17 @@
...
+# Script to import "email event" metrics from S3 into redshift.
+#
+
+import import_events
+SCHEMA = """
+    type VARCHAR(30) NOT NULL,
+    locale VARCHAR(40),
+    domain VARCHAR(40),
+    template VARCHAR(64),
+    flow_id VARCHAR(64)
```
philbooth: `flow_id` should be promoted to `DISTKEY` here, and it might be worth reading the redshift docs about `DISTKEY` and `SORTKEY` if you haven't already. They're pretty integral to query performance. (also, note that the `timestamp` column is automatically promoted to `SORTYKEY` inside `import_events.py`)

We probably also want to compress every column. @rfk ran some magic incantation on our other tables that said `ENCODE lzo` was the best setting elsewhere in this regard, so presumably it will be best here too.

You can see examples for both of the above in `import_flow_events.py`.

philbooth: > `flow_id` should be promoted to `DISTKEY` here...

Well, maybe it should. It depends on what the queries against this table will look like, but I presume you are going to join on `flow_id`.

import_email_events.py
```diff
@@ -0,0 +1,17 @@
...
+"""
+
+COLUMNS = "type, locale, domain, template, flow_id"
+
+import_events.run("fxa-email", "email", SCHEMA, COLUMNS, SCHEMA, COLUMNS)
```
philbooth: Subsequent to you opening this PR originally, I updated my code to use named arguments in these calls to `run`, to make the intent clearer (as there's so many args). See the bottom of e.g. `import_activity_events` for an example.

import_email_events.py
philbooth: Also, can `flow_id` be `NOT NULL` too, or are there occasions when it won't be present on the event?

rfk: IIUC some of these events will not have a flow_id, e.g. emails that we send related to "change password" which is not part of a flow.

rfk: One good rule of thumb I've found for choosing a distkey is, "if you were going to do a `COUNT DISTINCT` query on this table, which column would you be counting?".  I'm not sure there's an obvious answer here, but `flow_id` seems to come the closest.

vbudhram: > can flow_id be NOT NULL too

Yea the change/reset password emails might not have `flow_id` field. @rfk @philbooth I think the most obvious for distkey would be `flow_id`, ill update that.

philbooth:
@vbudhram, code looks good!

I'm taking the liberty of running an import now to try it out, hope that's cool with you. You're welcome to `DROP` those tables and re-import yourself later, if you want to get a feel for how it runs / what the logs look like / how long it takes / whatever.

And of course, if you have any questions about running the import jobs, just shout. (the readme for this repo has some useful info but I just noticed it needs to be brought up to date a bit, will do that at some point).

import_email_events.py
```diff
@@ -0,0 +1,22 @@
...
+    type VARCHAR(30) NOT NULL ENCODE lzo,
+    locale VARCHAR(40) ENCODE lzo,
+    domain VARCHAR(40) ENCODE lzo,
+    template VARCHAR(64) ENCODE lzo,
+    flow_id VARCHAR(64) DISTKEY ENCODE lzo
+"""
+
+COLUMNS = "type, locale, domain, template, flow_id"
+
+import_events.run(s3_prefix="fxa-email",
```
philbooth: The prefix needs to match as far as the hyphen immediately preceding the date in each file name, so this should be `"fxa-email/events"` in your case.

This fixes:

```
  File "./import_email_events.py", line 21, in <module>
    perm_columns=COLUMNS)
  File "/home/ec2-user/fxa-activity-metrics/import_events.py", line 225, in run
    unpopulated_days = get_unpopulated_days()
  File "/home/ec2-user/fxa-activity-metrics/import_events.py", line 161, in get_unpopulated_days
    if is_candidate_day(day) and not is_day_populated(day):
  File "/home/ec2-user/fxa-activity-metrics/import_events.py", line 148, in is_day_populated
    return bool(db.one(Q_CHECK_FOR_DAY.format(event_type=event_type, day=day)))
  File "/usr/lib/python2.7/site-packages/postgres/__init__.py", line 482, in one
    return cursor.one(sql, parameters, default)
  File "/usr/lib/python2.7/site-packages/postgres/cursors.py", line 105, in one
    out = self._some(sql, parameters, lo=0, hi=1)
  File "/usr/lib/python2.7/site-packages/postgres/cursors.py", line 127, in _some
    self.execute(sql, parameters)
psycopg2.DataError: invalid input syntax for type date: "events-2017-03-20"
```

philbooth:
> I'm taking the liberty of running an import now to try it out...

...aaaand it revealed a problem in `import_events.py`. I'll open a PR to fix that separately.

import_email_events.py
philbooth: Correction, it should be `"fxa-email/data/email-events"`, I only just had a look at S3. However that also isn't working properly yet, we might need to generalise `import_events.py` a bit more as part of this PR. Digging into it now...

vbudhram: Cool, I'll update accordingly

philbooth:
@vbudhram, an error message that you'll become quite familiar with if you spend any length of time importing data to redshift, is this one:

```
  File "import_email_events.py", line 21, in <module>
    perm_columns=COLUMNS)
  File "/home/ec2-user/fxa-activity-metrics/import_events.py", line 233, in run
    import_day(day)
  File "/home/ec2-user/fxa-activity-metrics/import_events.py", line 179, in import_day
    **CONFIG))
    cursor.run(sql, parameters)
    self.execute(sql, parameters)
psycopg2.InternalError: Load into table 'temporary_raw_email_data' failed.  Check 'stl_load_errors' system table for details.
```

I got this just now, trial-running the email event import.

When you encounter it, you want to query that table it mentions, `stl_load_errors`, to see what the problem is. The error information in there is pretty useful in my experience. You can either query it from the redshift console, or I also have a redash query set up that takes a date as the query param, e.g.:

https://sql.telemetry.mozilla.org/queries/1753/source?p_date=2017-04-13

(the date is the date the error occurred on, not the date of the file being imported)

In this case, the error we have is `Extra column(s) found` on the first line of `email-events-2017-04-12.csv`, which is the first file that the import tried to load (it runs in reverse chronological order).

So the problem here is that we need the columns in the `temp_schema` and `temp_columns` arguments to exactly match what is in the CSV.

Here, your code has six columns (five explicit plus the implied `timestamp` column), whereas the CSV contains eight. We need to define all eight columns in your script. Also, the order of `temp_columns` needs to exactly match the order of columns in the CSV. The easiest way to see all the CSV column names is in the lua script (it can be hard to find a line with every column populated in the CSV itself):

https://github.com/mozilla-services/puppet-config/blob/master/shared/modules/fxa/files/hekad/lua_outputs/fxa_email_csv.lua#L9-L16

So, all of that was a very long-winded way of saying we probably want `temp_columns` to be `"flow_id, domain, template, type, bounced, complaint, locale"` and then `temp_schema` needs to have the same columns (but not necessarily in the same order there).

Make sense?

philbooth:
And sorry, I realise I've failed to document any of that. When I update the readme here, all this will have to go in.

vbudhram:
@philbooth Nope that makes sense. I didn't include those fields because I didn't think they were needed in the table. Will update!

import_email_events.py
```diff
@@ -0,0 +1,24 @@
...
+    type VARCHAR(30) NOT NULL ENCODE lzo,
+    locale VARCHAR(40) ENCODE lzo,
+    domain VARCHAR(40) ENCODE lzo,
+    template VARCHAR(64) ENCODE lzo,
+    flow_id VARCHAR(64) DISTKEY ENCODE lzo,
+    complaint VARCHAR(64) ENCODE lzo,
+    bounced VARCHAR(64) ENCODE lzo
+"""
+
+COLUMNS = "type, locale, domain, template, flow_id, complaint, bounced"
```
philbooth: I think the order here is still not right? Looking at an example line from the CSV, `flow_id` should be first, `domain` should be second, `locale` should be last...

import_email_events.py
```diff
@@ -0,0 +1,24 @@
...
+
+import_events.run(s3_prefix="fxa-email/data/email-events",
+                  event_type="email",
+                  temp_schema=SCHEMA,
+                  temp_columns=COLUMNS,
+                  perm_schema=SCHEMA,
+                  perm_columns=COLUMNS)
```
philbooth: (also, I think you've already done this locally but, as per IRC, we'll need to add `id_column="flow_id"` to these arguments)

import_email_events.py
vbudhram: Thanks @philbooth! This was the cause of some strange errors I was experiencing. Data started importing once this was fixed.

vbudhram:
@philbooth Tested import script and wrote some queries for it, https://sql.telemetry.mozilla.org/queries/4236. Looks like it is working as expected. r?

import_email_events.py
```diff
@@ -0,0 +1,24 @@
...
+    flow_id VARCHAR(64) DISTKEY ENCODE lzo,
+    domain VARCHAR(40) ENCODE lzo,
+    template VARCHAR(64) ENCODE lzo,
+    type VARCHAR(64) NOT NULL ENCODE lzo,
+    complaint VARCHAR(64) ENCODE lzo,
+    bounced VARCHAR(64) ENCODE lzo,
+    locale VARCHAR(64) ENCODE lzo
+"""
+
+COLUMNS = "flow_id, domain, template, type, complaint, bounced, locale"
```
philbooth: I think `complaint` and `bounced` are still the wrong way round? Looks like the lua script processes them in that order:

https://github.com/mozilla-services/puppet-config/blob/master/shared/modules/fxa/files/hekad/lua_outputs/fxa_email_csv.lua#L40-L48

vbudhram:
@rfk Since phil is out mind an r? We would also have to hook this up to cron.

rfk (approved):
This looks OK to me, AFAICT all of @philbooth's comments have been addressed and it's working when run manually, let's get it hooked up for daily import.

rfk:
@vbudhram looks like the table schema changed when you swapped "bounce" and "complaint", should we drop the tables and re-import when this merges to master to ensure we get clean data?

vbudhram:
> should we drop the tables and re-import when this merges to master to ensure we get clean data

Done!

PR #71, opened 2017-04-13
fix(scripts): count from the end when parsing date from file name

philbooth:
The email events file name in S3 has an extra hyphenated part, which caused `import_events.py` to include an extra part in the date string. Instead of assuming that everything after the first hyphenated part is a date, we can fix it by counting back three parts from the end.

@vbudhram r?

### DIFF
--- a/import_events.py
+++ b/import_events.py
@@ -157,7 +157,9 @@ def run(s3_prefix, event_type, temp_schema, temp_columns, perm_schema, perm_colu
         print message
         for key in s3.list(prefix=s3_prefix):
             filename = path.basename(key.name)
-            day = "-".join(filename[:-4].split("-")[1:])
+            # Ignore the last four characters (".csv") then join the last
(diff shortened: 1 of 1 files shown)


import_events.py
```diff
@@ -157,7 +157,7 @@ def get_unpopulated_days():
         print message
         for key in s3.list(prefix=s3_prefix):
             filename = path.basename(key.name)
-            day = "-".join(filename[:-4].split("-")[1:])
+            day = "-".join(filename[:-4].split("-")[-3:])
```
vbudhram: >  we can fix it by counting back three parts from the end

It is not 100% obvious to me why go back 3 when only 1 extra hyphen was added?

vbudhram (approved):

import_events.py
philbooth: Good point, thanks. I've added a comment to make this clearer.

PR #76, opened 2017-04-26
feat(email): Import email events as part of `make import`

rfk:
This will cause them to be imported by the daily cronjob; @vbudhram r?


vbudhram:
Thanks! @rfk

Issue #73, opened 2017-04-14
Update redshift column encoding from lzo to zstd


philbooth:
While investigating #72, I noticed that we're not using the optimum column encoding on our tables any more. Apparently redshift added a new encoding in January, `zstd`. If the claimed reductions are are anything like the numbers below, we should switch to it.

Here's what I found:

```
fxa=# ANALYZE COMPRESSION flow_events;
    Table    |  Column   | Encoding | Est_reduction_pct
-------------+-----------+----------+-------------------
 flow_events | timestamp | zstd     | 58.17
 flow_events | type      | zstd     | 32.36
 flow_events | flow_id   | zstd     | 52.86
 flow_events | flow_time | zstd     | 23.66
 flow_events | locale    | zstd     | 91.98
 flow_events | uid       | zstd     | 91.98
(6 rows)

fxa=# ANALYZE COMPRESSION flow_metadata;
     Table     |    Column    | Encoding | Est_reduction_pct
---------------+--------------+----------+-------------------
 flow_metadata | flow_id      | zstd     | 46.65
 flow_metadata | begin_time   | zstd     | 59.74
 flow_metadata | duration     | zstd     | 19.37
 flow_metadata | completed    | zstd     | 16.32
 flow_metadata | new_account  | zstd     | 43.28
 flow_metadata | ua_browser   | zstd     | 45.54
 flow_metadata | ua_version   | zstd     | 49.29
 flow_metadata | ua_os        | zstd     | 42.58
 flow_metadata | context      | zstd     | 28.67
 flow_metadata | entrypoint   | zstd     | 45.76
 flow_metadata | migration    | zstd     | 88.63
 flow_metadata | service      | zstd     | 48.38
 flow_metadata | utm_campaign | zstd     | 50.26
 flow_metadata | utm_content  | zstd     | 41.69
 flow_metadata | utm_medium   | zstd     | 53.37
 flow_metadata | utm_source   | zstd     | 53.22
 flow_metadata | utm_term     | zstd     | 91.98
 flow_metadata | export_date  | zstd     | 91.54
 flow_metadata | locale       | zstd     | 91.98
 flow_metadata | uid          | zstd     | 91.98
(20 rows)

fxa=# ANALYZE COMPRESSION flow_experiments;
      Table       |   Column    | Encoding | Est_reduction_pct
------------------+-------------+----------+-------------------
 flow_experiments | experiment  | zstd     | 87.10
[...]

PR #78, opened 2017-05-04
fix(scripts): update column schemata for zstd encoding

philbooth:
Related to #73.


rfk (approved):
LGTM

philbooth:
These are in place now and appear to be working. Merging.

PR #82, opened 2017-06-07
feat(scripts): import counts

### LINKED ISSUE #68: Total FxA Accounts
As a product manager, I should be able to know how many accounts have been created to date.

We currently don't have an easy way to easily measure the total number accounts that have been created to date. 

As we work with marketing over 2017 to revive inactive FxA accounts, it would be nice to know how many accounts have ever been created.

It would also be nice to keep track of the proportion of active to active/inactive accounts.

As per how we roll-it up, we could have something like this (which doesn't load since it is too resource intensive in Re:Dash):
https://sql.telemetry.mozilla.org/queries/3916/source#7783

I'm open to other data formats though if we want more granularity.

philbooth:
Fixes #68.

Adds a new table to redshift, `counts`, that stores absolute daily counts of stuff. Currently that table includes two counts, `accounts` and `verified_accounts`. Perhaps it will grow more columns in the future, which is why I opted for such a generic table name.

You can see a redash chart that uses the new table [here](https://sql.telemetry.mozilla.org/queries/4783/source#9714) (actual numbers redacted because reasons):

Screen shot of a chart showing the total number of Firefox Accounts, with the numbers redacted

I implemented this as a standalone script rather than leaning on common code in `import_events.py`, for a couple of reasons:

1. Given that there is only 1 row of data per day, there's really no benefit to expiring old data. Better to offer the most complete history possible to queries.

2. The common code for importing events assumes a timestamp that's seconds-since-the-epoch, whereas the CSV for this contains YYYY-mm-dd. I could have refactored the common code to enable re-use here without numeric timestamps, or I could have asked @jbuck to change his export to write a timestamp. But I didn't see the point to be honest, the schema for this table seems simple enough to not bother leaning on anything else.

Note that these changes are currently running live on the redshift helper instance, because I wanted to see them work before opening the PR.

@vbudhram r?

### DIFF
--- a/import_counts.py
+++ b/import_counts.py
@@ -1,0 +1,103 @@
+from os import path
+from datetime import datetime
+import json
+import boto.s3
+import boto.provider
+import postgres
+
+with open("config.json") as f:
+    CONFIG = json.loads(f.read())
+
+if "aws_access_key_id" not in CONFIG:
+    aws = boto.provider.Provider("aws")
+    CONFIG["aws_access_key_id"] = aws.get_access_key()
+    CONFIG["aws_secret_access_key"] = aws.get_secret_key()
+
+S3_REGION = "us-east-1"
+S3_BUCKET = "net-mozaws-prod-us-west-2-pipeline-analysis"
+S3_PREFIX = "fxa-basic-metrics/"
+S3_URI = "s3://" + S3_BUCKET + "/" + S3_PREFIX + "fxa-basic-metrics-{day}.txt"
+DB_URI = "postgresql://{db_username}:{db_password}@{db_host}:{db_port}/{db_name}".format(**CONFIG)
+COUNTS_BEGIN = datetime.strptime("2017-05-30", "%Y-%m-%d")
+
+Q_DROP_CSV_TABLE = "DROP TABLE IF EXISTS temporary_raw_counts;"
+
+Q_CREATE_COUNTS_TABLE = """
+    CREATE TABLE IF NOT EXISTS counts (
+      day DATE NOT NULL UNIQUE SORTKEY ENCODE RAW,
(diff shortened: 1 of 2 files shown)


import_counts.py
```diff
@@ -0,0 +1,103 @@
...
+S3_URI = "s3://" + S3_BUCKET + "/" + S3_PREFIX + "fxa-basic-metrics-{day}.txt"
+DB_URI = "postgresql://{db_username}:{db_password}@{db_host}:{db_port}/{db_name}".format(**CONFIG)
+COUNTS_BEGIN = datetime.strptime("2017-05-30", "%Y-%m-%d")
+
+Q_DROP_CSV_TABLE = "DROP TABLE IF EXISTS temporary_raw_counts;"
+
+Q_CREATE_COUNTS_TABLE = """
+    CREATE TABLE IF NOT EXISTS counts (
+      day DATE NOT NULL UNIQUE SORTKEY ENCODE RAW,
+      accounts BIGINT NOT NULL ENCODE ZSTD,
```
philbooth, line 28: `BIGINT` for when we edge past the 4-billion-accounts mark. :smile:

import_counts.py
```diff
@@ -0,0 +1,103 @@
...
+"""
+
+Q_CHECK_FOR_DAY = """
+    SELECT day FROM counts
+    WHERE day = '{day}';
+"""
+
+Q_CREATE_CSV_TABLE = """
+    CREATE TABLE IF NOT EXISTS temporary_raw_counts (
+      day CHAR(10) NOT NULL UNIQUE SORTKEY,
```
philbooth, line 40: ...although perhaps I should also have made this `VARCHAR(11)` so that we don't need to change the script in 8,000 years time. :smile:

vbudhram (approved):
@philbooth This LGTM 👍

PR #83, opened 2017-06-16
feat(schema): add a continued_from column to flow_metadata

philbooth:
Related to mozilla/fxa-auth-server#1945.

This change just performs similar update logic to the other post-import updates we run on the `flow_metadata` table, setting a `continued_from` column based on the `flow.continued.${flow_id}` event that was added in mozilla/fxa-auth-server#1946. Hopefully there's nothing surprising here.

As an initial test, I:

1. Added a `continued_from` column to `flow_metadata` (and its sampled brethren).
2. Changed the `type` column on `flow_events` (and siblings) to be `VARCHAR(79)`.
3. Deleted the flow data for `2017-06-15`.
4. Re-imported `2017-06-15` using the updated script from this PR.

The import finished okay and the data seems fine. Obviously the real test will come when these events start showing up in the data. I left the updated script in place on the redshift helper so it should just start working as and when.

@vbudhram, @rfk r?

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -4,8 +4,10 @@
 
 import import_events
 
+# flow_id is VARCHAR(64) because it's 32 bytes hex-encoded
+# type is VARCHAR(79) so it can contain `flow.continued.${flow_id}`
 TEMPORARY_SCHEMA = """
-    type VARCHAR(64) NOT NULL ENCODE zstd,
+    type VARCHAR(79) NOT NULL ENCODE zstd,
     flow_id VARCHAR(64) NOT NULL DISTKEY ENCODE zstd,
     flow_time BIGINT NOT NULL ENCODE zstd,
     ua_browser VARCHAR(40) ENCODE zstd,
@@ -45,7 +47,7 @@ TEMPORARY_COLUMNS = """
 """
 
 EVENT_SCHEMA = """
-    type VARCHAR(64) NOT NULL ENCODE zstd,
+    type VARCHAR(79) NOT NULL ENCODE zstd,
     flow_id VARCHAR(64) NOT NULL DISTKEY ENCODE zstd,
     flow_time BIGINT NOT NULL ENCODE zstd,
     locale VARCHAR(40) ENCODE zstd,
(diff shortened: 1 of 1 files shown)


import_flow_events.py
```diff
@@ -5,7 +5,7 @@
 import import_events
 
 TEMPORARY_SCHEMA = """
-    type VARCHAR(64) NOT NULL ENCODE zstd,
+    type VARCHAR(79) NOT NULL ENCODE zstd,
```
vbudhram: Should this be length 80? `continue:flow_id` has 16 characters. I think it might be worthwhile to add a quick comment about it.

vbudhram (approved):
@philbooth Just the one question, otherwise LGTM 👍

import_flow_events.py
philbooth: I've added a comment.

The event is `flow.continued.${flow_id}` so 15 + 64 = 79 characters. I briefly considered making it 80 characters on account of that being a round number, but it's only a round number if you're a ten-fingered weirdo from planet earth. Whereas 79 is prime in the universal language of maths, which makes it a round number EVERYWHERE. :smile:

philbooth:
Thanks for the review @vbudhram, I'll hold off merging until we actually know for sure it is working against the new events.

rfk (approved):
lgtm!

vladikoff:
from mtg: Ready to merge, but needs to check if this is working first

vladikoff:
from mtg: stuff that uses this has not shipped yet

vladikoff:
from mtg: progress being made this train (train-93)

philbooth:
This has kicked in now, you can see it used for the phase 3 numbers in this dashboard:

https://sql.telemetry.mozilla.org/dashboard/connect-another-device-phase-2

philbooth:
> ...you can see it used for the phase 3 numbers in this dashboard...

Err, no you can't. I misremembered what that query actually does. This one shows the `continued_from` stuff though:

https://sql.telemetry.mozilla.org/queries/20538/source#52717

PR #84, opened 2017-07-10
feat(schema): add a stricter definition of multi-device users

philbooth:
This is an experimental new definition of the `daily_multi_device_users` table, which tries to discount the effect of zombie sessions causing single-device users to show up as multi-device. It tweaks the `INSERT` logic by requiring that the `present` device is used both before and after the `past` device, thus indicating that the two devices have been in use simultaneously. The tweak is applied against a new table rather than the existing one, so that we don't mess with any existing queries and can compare the two definitions side-by-side.

I copied this version of the script over to the redshift helper on Friday, so we have a couple of days history in the table already and it is gradually filling up day-by-day. I did also run a script to initialise the new table with a full 3 months of old data, but the query died running out of disk space. If it's important to have that data immediately, I could set the initial import to run on a day-by-day basis instead. But since this is only experimental anyway, I figured it was fine to let it populate gradually and we can see how it goes over time.

@vbudhram, super low-priority r?

### DIFF
--- a/calculate_daily_summary.py
+++ b/calculate_daily_summary.py
@@ -129,48 +129,92 @@ Q_VACUUM_TABLES = """
     ANALYZE daily_multi_device_users{suffix};
 """
 
+# The "strict" multi-device users table tweaks the multi-device classification
+# to ensure that the the current device was also used at some point before the
+# previous device.
+Q_STRICT_MD_USERS_CREATE_TABLE = """
+    CREATE TABLE IF NOT EXISTS strict_daily_multi_device_users (
+      day DATE NOT NULL SORTKEY ENCODE RAW,
+      uid VARCHAR(64) NOT NULL DISTKEY ENCODE zstd,
+      device_now VARCHAR(32) NOT NULL ENCODE zstd,
+      device_prev VARCHAR(32) NOT NULL ENCODE zstd
+    );
+"""
+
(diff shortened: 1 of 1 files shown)


calculate_daily_summary.py
```diff
@@ -129,48 +129,92 @@
...
+Q_VACUUM_STRICT_MD_USERS = """
+    END;
+    VACUUM FULL strict_daily_multi_device_users;
+    ANALYZE strict_daily_multi_device_users;
+"""
+
 def summarize_events():
     db = postgres.Postgres(DB)
-    db.run("BEGIN TRANSACTION")
-    try:
```
philbooth, line 135: I ditched the transaction stuff to match our other import scripts. I'd observed that this wasn't actually running queries in a transaction, so failures were just leaving the database in whatever broken state it failed at. I'm sure it's possible to fix that properly somehow, but occasionally needing to drop a day of data and then re-run the script hasn't been an annoying enough process for me to bother looking into it yet.

vbudhram (approved):
@philbooth LGTM 👍

philbooth:
This seems to be working fine, you can see the new data in this chart:

https://sql.telemetry.mozilla.org/queries/4776#9696

Seems to be fairly steady around 70k below the non-strict multi-device user count.

Issue #99, opened 2018-02-07
Prepare for mozlog 2


philbooth:
As part of the move to mozlog 2, we want to make some changes to these import scripts:

* Change the entry point to handle Lambda message objects.
* Add whatever concurrency is needed to keep Lambda happy.
* Stop assuming that one day is the atomic payload size (so Redshift updates at the same frequency as Amplitude).

While doing that, we may or may not port them to node, depending on how stuff pans out.

Once it's ready, we plan to run both pipelines side-by-side against the current mozlog 1 format. Only when we're happy that they are the same will we flip the code in the content server over to mozlog 2.

jbuck:
After a quick look at the [Kinesis Firehose docs](https://docs.aws.amazon.com/firehose/latest/dev/create-destination.html#create-destination-redshift), I don't think we should actually make the first two changes on this list:

> * Change the entry point to handle Lambda message objects.
> * Add whatever concurrency is needed to keep Lambda happy.
> * Stop assuming that one day is the atomic payload size (so Redshift updates at the same frequency as Amplitude).

Basically, Kinesis Firehose will save to S3, then run the redshift COPY to a single table. All we should need to do is modify the scripts so that they can work with a single staging table, and then we're good!

philbooth:
That's great, thanks @jbuck!

philbooth:
> > * Stop assuming that one day is the atomic payload size (so Redshift updates at the same frequency as Amplitude).
>
> Basically, Kinesis Firehose will save to S3, then run the redshift COPY to a single table. All we should need to do is modify the scripts so that they can work with a single staging table, and then we're good!

Fwiw, I made a start on this reduced-functionality script in #100, see `kinesis_flow_events_2.py` in that PR. Format/structure will probably need to change for actual integration with Kinesis, but this works for running from the command line. See `kinesis_flow_events_1.py` in the same PR for the schema of the staging table and the CSV file.

philbooth:
Adding @jbuck's face to this to carry on with while I'm away. Related flow import PR is in #100.

vladikoff:
from mtg: come back to this in 110

philbooth:
IIUC from last night's meeting, this probably shouldn't be in `next` any more. Moving to `backlog`.

vladikoff:
Let's reopen if this comes up again

PR #100, opened 2018-02-20
wip: break up the flow import for kinesis integration

philbooth:
Related to #99. Not for merging, just for discussion / ongoing work.

I'm opening this for when @jbuck gets back from holiday, hopefully it is sufficient to unblock him on the mozlog 2 stuff. (I disappear off on a holiday of my own shortly after he gets back)

What I've done here is break apart the flow import into 2 separate scripts:

* `kinesis_flow_events_1.py`, which you can mostly ignore. This one is just here so I can set up and populate our temporary / raw data table for the other script to read from. The work in this script will ultimately be done by Kinesis Firehose, when he have the real mozlog 2 pipeline up and running. It can also serve as a reference when setting up the Kinesis stuff, for what the Redshift schema needs to be and what the expected CSV format is.

* `kinesis_flow_events_2.py`, which contains the good stuff. This script reads data from the temporary / raw data table and makes no assumptions about the length of time that table covers. For more about why we want to do this, see the discussion in #99.

Something that we lose with these changes is straightforward error correction. Because the existing scripts do everything in atomic units of a day, it's easy to delete and re-import specific days in the event of something going wrong. But hopefully our shiny new pipeline will be so perfect that we don't have to worry about that too much. (this is why there are no `export_date` columns in the new schema)

The other side-effect of the loss of per-day semantics is that we now populate `flow_metadata` first, so that we can use `MAX(begin_time)` when drawing the line for data expiry. That seemed like the right thing to do, although maybe it's not that important really.

If anyone wants to muck about with these scripts for real, they're in `~/kinesis-flow-events` on our redshift helper EC2 instance. There's also a week of data currently imported to `kinesis_` prefixed table names in Redshift, which I imported (and tested) just to make sure the code actually works.

### DIFF
--- a/kinesis_flow_events_1.py
+++ b/kinesis_flow_events_1.py
@@ -1,0 +1,110 @@
+#
+# Script to copy flow metrics from CSV to a temporary table
+#
+
+import json
+import boto.s3
+import boto.provider
+import postgres
+
+with open("config.json") as f:
+  CONFIG = json.loads(f.read())
+
+if "aws_access_key_id" not in CONFIG:
+  aws = boto.provider.Provider("aws")
+  CONFIG["aws_access_key_id"] = aws.get_access_key()
+  CONFIG["aws_secret_access_key"] = aws.get_secret_key()
+
+S3_BUCKET = "net-mozaws-prod-us-west-2-pipeline-analysis"
+S3_URI = "s3://" + S3_BUCKET + "/fxa-flow/data/flow-{day}.csv"
+DB_URI = "postgresql://{db_username}:{db_password}@{db_host}:{db_port}/{db_name}".format(**CONFIG)
+
+TABLE_NAME = "kinesis_temporary_raw_flow_data"
+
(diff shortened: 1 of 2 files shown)


philbooth:
Note that when we come to doing this for real, we'll need similar scripts for the activity and email events too. I figured it made sense to work everything out for one import first, and the flow import is the most complicated so seemed like the best test-candidate.

philbooth:
Tangential fix alert.

While doing some other analysis I noticed that our import scripts currently store all of the raw `flow.continued.{flow_id}` events in the `flow_events` table. As well as wasting space, it makes the result set of `SELECT DISTINCT type FROM flow_events` really, really unwieldy.

Fixing it is easier in this PR's codebase than it is in `master` because of all the refactoring that's happened over there, so if we're ultimately ditching those scripts anyway I didn't see the point of doing it in `master`.

Hence 273a56f.

vladikoff:
from mtg: closing

Issue #104, opened 2018-08-15
Debigulate the data


philbooth:
There's a significant proportion of redundant data in our current dataset. We could:

* Delete all the `flow.continued...` events after using them to set the `continued_from` column in `flow_metadata`.

* Delete all the `flow.experiment...` events after using them to populate the `flow_experiments` table.

* Delete all the `flow.begin` and `flow.completed` events after using them to populate `flow_metadata` (be careful with this one, it might break some queries).

* Delete the `strict_multi_device_users` import job, which takes aaages and sometimes fails to finish at all.

* Send less performance events, just keep the ones we're actually using.

I estimate we could save about 5 gigabytes from our ~7 gigabyte per-day dataset size.

/cc @jbuck

shane-tomlinson:
> I estimate we could save about 5 gigabytes from our ~7 gigabyte per-day dataset size.

Crikey. I knew we were operating at some scale, didn't realize the logs alone were this much.

philbooth:
> * Delete all the `flow.begin` and `flow.completed` events after using them to populate `flow_metadata` (be careful with this one, it might break some queries).

I just checked and this one will break a number of queries. They could all be fixed to use `flow_metadata` columns instead, but that's probably not a fair burden to put on people.

Although, if a query breaks and nobody notices it's broken, does it matter?

Not sure, I'll send an email to canvas opinion.

irrationalagent:
I'm going to weigh in on this soon. I know I've used flow.completed a fair bit, just need to see where.

irrationalagent:
So there are a few charts on [this dashboard](https://sql.telemetry.mozilla.org/dashboard/fxa-big-picture-re-dash-edition) that will break bcs of `flow.completed` use, as well as queries that aren't on there but are forks of those. Given the huge benefit in terms of space here though, I would be OK with the change. I can just patch those queries to join on `flow.metadata`. I haven't used `flow.begin` much, I usually use `.view` as top-of-funnel.

philbooth:
@irrationalagent, don't worry about `flow.completed`, I'm going to keep it. The goal of this issue wasn't to cause upheaval, it was to get whichever speedups could be had without breaking stuff. :smile:

`flow.begin` is toast though!

PR #105, opened 2018-08-17
Debigulate flow_events

philbooth:
Related to #104.

This includes the three less contentious changes to reduce the amount of data we store in Redshift. They don't break any queries as far as I can make out.

@jbuck r?

(if you think things look okay from a syntax standpoint I'll do a trial run import before merging, just to make sure everything still works)

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -248,6 +248,12 @@ Q_UPDATE_CONTINUED_FROM = """
     WHERE flow_metadata{suffix}.flow_id = continued.flow_id;
 """
 
+Q_DELETE_CONTINUED_EVENTS = """
+    DELETE FROM {table_name}
+    WHERE timestamp::DATE <= '{day}'
+    AND type LIKE 'flow.continued.%';
+"""
+
 Q_INSERT_EXPERIMENTS = """
     INSERT INTO flow_experiments{suffix} (
       experiment,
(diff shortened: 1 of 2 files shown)


import_flow_events.py
```diff
@@ -248,6 +248,12 @@
     WHERE flow_metadata{suffix}.flow_id = continued.flow_id;
 """
 
+Q_DELETE_CONTINUED_EVENTS = """
+    DELETE FROM {table_name}
+    WHERE timestamp::DATE <= '{day}
```
jbuck: Missing a `'` here

import_flow_events.py
```diff
@@ -285,6 +291,12 @@
     WHERE flow_experiments{suffix}.flow_id = events.flow_id;
 """
 
+Q_DELETE_EXPERIMENT_EVENTS = """
+    DELETE FROM {table_name}
+    WHERE timestamp::DATE <= '{day}
```
jbuck: Missing a `'` here

jbuck (approved):

import_flow_events.py
philbooth: Doh! Thanks!

philbooth:
These landed:

```
fxa=# select count(*) from flow_events where type like 'flow.continued%';
 count
-------
     0
(1 row)

fxa=# select count(*) from flow_events where type like 'flow.experiment%';
 count
-------
     0
(1 row)

fxa=# drop table strict_daily_multi_device_users;
DROP TABLE
```

PR #106, opened 2018-08-28
feat(scripts): delete flow.begin events after processing

### LINKED ISSUE #104: Debigulate the data

philbooth:
Fixes #104.

This is the last thing I'm going to do about reducing the size of our data. `flow.completed` survives. Have pushed to the redshift helper for tomorrow morning's import, to check it's okay before merging.

@mozilla/fxa-devs

### DIFF
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -145,7 +145,7 @@ Q_INSERT_METADATA = """
       FROM {table_name}
     )
     WHERE sample <= {percent}
-    AND type LIKE 'flow%begin';
+    AND type = 'flow.begin';
 """
 
 Q_UPDATE_METADATA = """
--- a/import_flow_events.py
+++ b/import_flow_events.py
@@ -148,6 +148,12 @@ Q_INSERT_METADATA = """
     AND type = 'flow.begin';
 """
 
+Q_DELETE_BEGIN_EVENTS = """
+    DELETE FROM {table_name}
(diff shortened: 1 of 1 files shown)


philbooth:
```
fxa=# select count(*) from flow_events where type = 'flow.begin';
 count
-------
     0
(1 row)
```

vladikoff (approved):

vladikoff:
@philbooth looks good, do we want final confirmation from @davismtl or @irrationalagent ?

philbooth:
> do we want final confirmation from @davismtl or @irrationalagent ?

Well it's a bit late for that if we do, as I already deleted all the events. 😄

irrationalagent:
for the record this was what we had agreed to do, so I'm fine with it.

PR #116, opened 2018-12-12
feat(scripts): dockerize the import/summarize scripts

### LINKED ISSUE #115: Dockerize these scripts so we can run them from the fxa admin node
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1513129#c5

philbooth:
Fixes #115.

WIP. Don't review yet, I'm only PRing so I can work through build issues in Circle...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,54 @@
+version: 2
+
+jobs:
+  build:
+    docker:
+      - image: circleci/python:2.7
+
+    working_directory: /dockerflow
+
+    steps:
+      - checkout
+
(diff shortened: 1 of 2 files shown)


vladikoff:
Nice!

philbooth:
Hmm, Circle not even starting the builds, I wonder if it's because `.circleci` is not in master. Closing this and pushing straight there instead....

philbooth:
I'm an idiot, I forgot I'd put the filter in to only build tags. Oh well, it's already in `master` now, I'll continue force-pushing there...

PR #117, opened 2018-12-13
fix(docker): Fix image tag

jbuck:
I added the creds to CircleCI but the build failed because `fxa-email-service:build` is in the wrong repo. I think after this patch it should push up to Dockerhub correctly

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ jobs:
 
       - run:
           name: Build docker image
-          command: docker build -f Dockerfile -t fxa-activity-metrics .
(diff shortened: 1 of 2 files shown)


philbooth:
Sorry @jbuck, didn't see this.

I've been force-pushing to `master` because I thought I was the only person working in this repo and pull requests are for chumps :scream:. So now you have merge conflicts.

But I think the thing you were fixing got fixed amongst it all anyway, @rfk pointed it out to me this morning too.

jbuck:
Yep, looks like it's fixed now!

PR #118, opened 2018-12-13
fix(s3): Connect to correct S3 bucket region

jbuck:
The S3 bucket is in the us-west-2 region. Boto is able to follow redirects to the right place, but this means no redirect needed


philbooth (approved):
Oh, thanks! :+1:

vladikoff:
👍

PR #119, opened 2018-12-14
feat(credentials): Support using an IAM role for credentials.

rfk:
Here's my quick pitch for supporting an `AWS_IAM_ROLE` env var, per https://github.com/mozilla/fxa-activity-metrics/issues/115#issuecomment-447138271.  It runs locally and appears to spit out the right strings but I haven't tried actually connecting it to a redshift cluster...

### DIFF
--- a/import_activity_events_retro.py
+++ b/import_activity_events_retro.py
@@ -35,8 +35,16 @@ def env_or_default(variable_name, default_value):
     return default_value
 
 p = boto.provider.Provider("aws")
-AWS_ACCESS_KEY = env_or_default("AWS_ACCESS_KEY", p.get_access_key())
-AWS_SECRET_KEY = env_or_default("AWS_SECRET_KEY", p.get_secret_key())
+AWS_IAM_ROLE = env_or_default('AWS_IAM_ROLE', None)
+if AWS_IAM_ROLE is not None:
+    CREDENTIALS="aws_iam_role={AWS_IAM_ROLE}".format(AWS_IAM_ROLE=AWS_IAM_ROLE)
+else:
+    AWS_ACCESS_KEY = env_or_default("AWS_ACCESS_KEY", p.get_access_key())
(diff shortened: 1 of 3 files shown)


import_activity_events_retro.py
```diff
@@ -35,8 +35,16 @@ def env_or_default(variable_name, default_value):
     return default_value
 
 p = boto.provider.Provider("aws")
-AWS_ACCESS_KEY = env_or_default("AWS_ACCESS_KEY", p.get_access_key())
-AWS_SECRET_KEY = env_or_default("AWS_SECRET_KEY", p.get_secret_key())
+AWS_IAM_ROLE = env_or_default('AWS_IAM_ROLE', None)
```
philbooth, line 38: Btw (tangent, sorry), it wasn't clear to me whether it was worth the effort to still maintain these old `_retro` versions of the import scripts. I *think* they were written when we wanted to import data from an older format to the current schema, but it's a very vague memory and I might be wrong about that. They've certainly not been used in the last year.

Maybe we should just delete them, what do you think?

import_counts.py
```diff
@@ -24,8 +24,16 @@ def env_or_default(variable_name, default_value):
     return default_value
 
 aws = boto.provider.Provider("aws")
```
philbooth, line 26: No big deal but this could move into the `else:` branch too.

philbooth (approved):
Lgtm! :+1:

import_activity_events_retro.py
philbooth: (talking to myself now) `git log` points at 5b0f1dc639b0ee77fd160924bd55cfaa011f6da6, which comes from [this PR](https://github.com/mozilla/fxa-activity-metrics/pull/17). So I think we're good to delete these `_retro` import scripts, it's been over 2 years now and nobody is interested in the old data any more.

PR #122, opened 2018-12-18
fix(scripts): Allow a small number of import errors

jbuck:


import_counts.py
```diff
@@ -75,6 +75,7 @@ def env_or_default(variable_name, default_value):
     FROM '{s3_uri}'
     CREDENTIALS '{CREDENTIALS}'
     FORMAT AS CSV
+    MAXERROR AS 10
```
philbooth, line 78: Fwiw, I'd be comfortable if this was at least an order of magnitude greater, if it potentially prevents a re-import in the future. 100 events in the grand scheme of things is a drop in the ocean, whereas 10 seems perilously close to the 4 you just experienced.

philbooth (approved):
LGTM, although I'd probably go further than 10...

philbooth (approved):

PR #123, opened 2018-12-28
feat(scripts): add country and region columns to activity/flow events

### LINKED ISSUE #108: Store country in our own data
We currently pass country to Amplitude but we don't store it in our own dataset for viewing in Re:Dash.

Let's keep our own hard copy of country for use in Re:Dash.

Story:
As a Firefox leader, I would like to be able to track changes in trends in our most important countries. (e.g. YoY)

philbooth:
Fixes #108.

Adds 2 extra columns, `country` and `region`, to the `activty_events` and `flow_events` tables. We need to get those 2 columns added to the CSV before we can apply this change (hence `WIP`), because without them the import will fail. This also means that, once applied, the scripts won't be able to re-import old data from before the columns existed.

@jbuck, based on the discussion in https://github.com/mozilla/fxa-activity-metrics/issues/108#issuecomment-450213718, do you think stackdriver logging will happen soon enough for us to just add columns to the CSV that way, or do you think we'll need to update the Heka filter as an initial step? I suspect it's the latter, but worth asking.

And if we do need to update the Heka filter, do you know who we should ask to help us? I used to know how to do this, but I can't remember what was involved now, or even where to find the code.

### DIFF
--- a/import_activity_events.py
+++ b/import_activity_events.py
@@ -11,10 +11,12 @@ SCHEMA = """
     service VARCHAR(40) ENCODE zstd,
     ua_browser VARCHAR(40) ENCODE zstd,
     ua_version VARCHAR(40) ENCODE zstd,
-    ua_os VARCHAR(40) ENCODE zstd
+    ua_os VARCHAR(40) ENCODE zstd,
+    country VARCHAR(40) ENCODE zstd,
+    region VARCHAR(40) ENCODE zstd
 """
 
-COLUMNS = "ua_browser, ua_version, ua_os, uid, type, service, device_id"
+COLUMNS = "ua_browser, ua_version, ua_os, uid, type, service, device_id, country, region"
 
 import_events.run(s3_prefix="fxa-retention/data/events",
                   event_type="activity",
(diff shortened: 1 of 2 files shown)


import_activity_events.py
```diff
@@ -11,10 +11,12 @@
     service VARCHAR(40) ENCODE zstd,
     ua_browser VARCHAR(40) ENCODE zstd,
     ua_version VARCHAR(40) ENCODE zstd,
-    ua_os VARCHAR(40) ENCODE zstd
+    ua_os VARCHAR(40) ENCODE zstd,
+    country VARCHAR(40) ENCODE zstd,
+    region VARCHAR(40) ENCODE zstd
```
vladikoff, line 16: do we need to worry about anything else besides `county` and `region` fields, is there more data we get from the geo response?

vladikoff (approved):
seems sane, flagged @jbuck to also review

jbuck:
Yes, I think we'll have Stackdriver logging in place in early January

philbooth:
> Yes, I think we'll have Stackdriver logging in place in early January

This is brilliant news!

jbuck:
@philbooth will I need to run a manual DML change to add these two columns to the appropriate tables?

jbuck:
Also, can we merge this now? or will the additional columns not being present in the current data set cause problems?

philbooth:
> will I need to run a manual DML change to add these two columns to the appropriate tables?

Yep, that's what I was thinking.

> Also, can we merge this now? or will the additional columns not being present in the current data set cause problems?

I'm pretty sure the missing columns will cause it to fail. ISTR trying that before, a long time ago.

PR #128, opened 2019-02-11
feat(docs): update the readme with details of the new data source

philbooth:
The readme in this repo had gone very stale. I've deleted most of it, because all the stuff about the temporary data source and so on is out of date now. I've mentioned the new data source instead and also that there is a migration away from Heka in progress.

@mozilla/fxa-devs r?


vladikoff (approved):
lgtm!

PR #131, opened 2019-02-20
fix(scripts): fix off-by-one errors in metrics sample rates

### LINKED ISSUE #129: Sampled datasets appear to be closer to 11% and 51%
Got a request for data going back as far as possible. Using some cached results in redshift its actually possible to cobble together a time series of sampled data stretching all the way back to 2015.

Unfortunately, when comparing our unsampled data to our sampled datasets I was reminded of the fact that the sampled versions, when multiplied by the appropriate factor (x10 or x2), overestimate the data by a [very consistent factor](https://sql.telemetry.mozilla.org/queries/61398#158128). I neglected to file a bug earlier so here I am.

The business part of sampling appears to be done around here: https://github.com/mozilla/fxa-activity-metrics/blob/4141223dfab61e4c34fb18204a6219800f41903b/import_events.py#L113-L117

Is the problem here that there is a 0 cohort that gets included in the comparison above, e.g. you end up with cohorts 0-10 or 0-50, which is 11 cohorts? In other words, line 117 should really be `WHERE cohort < {percent}`.

I'm filing this issue but I'm not sure we should fix it because it would cause some data weirdness. Maybe we should just rename the datasets to `*_sampled_11` and `*_sampled_51`. You would then need to multiply the results by approximately 9.0909 and 1.9608, respectively, if you want a more accurate estimate of the true volume. 

CC @davismtl I believe this to be one source of the mismatch between re:dash and amplitude that we've observed when trying to get a long-run sense of YoY changes.

philbooth:
Fixes #129 (well, along with a re-import of data it does).

@jbuck, @irrationalagent r?

### DIFF
--- a/import_events.py
+++ b/import_events.py
@@ -114,7 +114,7 @@ Q_INSERT_EVENTS = """
             STRTOL(SUBSTRING({id_column} FROM 0 FOR 8), 16) % 100 AS cohort
         FROM {temp_table}
     )
-    WHERE cohort <= {percent}
+    WHERE cohort < {percent}
     AND ts::DATE = '{day}'::DATE
     AND ts::DATE >= '{max_day}'::DATE - '{months} months'::INTERVAL;
 """.format(perm_table=TABLE_NAMES["perm"],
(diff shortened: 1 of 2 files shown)


philbooth:
(@irrationalagent, you'll notice from the diff that this sampling code got copy-pasted to a few places, so presumably other metrics derived from it were wrong too 😊)

irrationalagent (approved):
Looks like you got all of them, thanks!

